PR #84, opened 2019-08-15
Prevent UI tests from failing randomly (#81)

### LINKED ISSUE #81: UI tests randomly fail
When executed, some UI tests can fail unexpectedly. The following failures have been observed:

- the build enter an infinite loop,
- some assertions fail,
- unexpected exceptions are failed.

I believe that SWTBot is not always used properly, e.g. tests do not always wait for shells to close before moving on.

echebbi:
Addresses issue #81 _UI tests randomly fail_

Run a dozen of builds on both Travis CI and my own computer; none of them failed. Hence, it seems that issue is fixed. However, even if the build is way more stable now, some unexpected failures may still show up. I'll wait a bit before merging the PR. 

Here are the new improvements:

- wait for the "New Class" wizard to close before moving on,
- when testing the _PIT Mutations_ view make sure that the event triggered by double-clicking on a mutation has been fully processed before checking its effects
- try to fail fast when a compilation error is detected instead of waiting Travis CI to kill the build after 10 min

### DIFF
--- a/bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/mutations/OpenMutationDoubleClick.java
+++ b/bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/mutations/OpenMutationDoubleClick.java
@@ -16,9 +16,6 @@
 
 package org.pitest.pitclipse.ui.view.mutations;
 
-import com.google.common.base.Function;
-import com.google.common.base.Optional;
-
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IWorkspaceRoot;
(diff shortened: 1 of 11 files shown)


LorenzoBettini:
IMHO there are two many System.out.println now ;)

echebbi:
Yes, this is also my opinion. I'm about to squash all these commits and remove useless logs before merging the PR. 

For the record, 1 test still randomly fails (hopefully it's the last one). If I remember well it's [this one](https://github.com/pitest/pitclipse/blob/06fa7cd831ec42483bf0f5c247fb13bcdd5325db/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/j04_multiple%20projects_in_workspace.feature#L76), when we attempt to run mutation testing on `project4`. The last log is [here](https://travis-ci.com/pitest/pitclipse/builds/130587525#L10363). Basically, the build enters an infinite loop because of unexpected compilation errors (within projects programmatically created). Indeed, because of them an error popup is shown when SWTBot attempts to run the Pitest launch configuration. Since this popup is not expected we do not handle it and the test remains stuck there.

I don't know yet why these compilation errors occur, I guess there's some concurrency involved somewhere. I think this could be solved as a side effect of addressing #85.

echebbi:
Build still fail from time to time but it seems to be way more reliable (I'd said that about 1 build in 10 fails). Let's merge now and see how it goes.

@LorenzoBettini this should ease future contributions.

PR #88, opened 2019-10-10
Bump Pitest version to 1.4.10

echebbi:
> ***WIP:** For an unknown reason upgrading PIT lead to several runtime errors. Should take the time to investigate this.*

Changelog since 1.4.6:

1.4.10
 - Smaller blocks for more precise test targeting (thanks @jon-bell)
 - Fix A0D2 map key (thanks @Vampire)
 - Escape characters in init methods for html report (thanks @Vampire)
 - Filter junk mutations to compiler generated Objects.requireNonNull calls

1.4.9
 - Fix for powermock issues on (thanks @jon-bell)
 - Improved error message when no test plugin (thanks @szpak)
 - Support annotation processors such as Micronaut that do not set debug filename

1.4.8
 - Fix for bug in coverage when large number of classes (thanks @jon-bell)
 - Avoid stealing keyboard focus on macos (thanks @maxgabut)

1.4.7
 - Faster coverage calculation (thanks @jon-bell)


echebbi:
Replace by PR #107

PR #90, opened 2019-11-23
Upgrade Tycho to 1.5.1

Bananeweizen:
Since a released version is available, pitclipse should no longer use a
snapshot for builds.


echebbi:
Thanks for the update, I didn't have time to do it! I'll merge it as soon as the CI is happy; I may have to relaunch the build because of #81.

PR #92, opened 2020-02-05
Task 91 - Log4j2 packages should be excluded by default

### LINKED ISSUE #91: Log4j2 packages should be excluded by default
## Motivation

Currently, the launch configuration excludes by default the following packages `java.util.logging, org.apache.log4j, org.slf4j, org.apache.commons.logging` but in Log4j2 the package for logging is different: `org.apache.logging.log4j`, and you get survived mutants when removing Log4j2 logging statements.

## Proposed Solution

`org.apache.logging.log4j` should be excluded as well by default.

I'll try to provide a PR

LorenzoBettini:
Closes #91 

https://github.com/pitest/pitclipse/issues/91


echebbi:
Thanks for the update!

PR #93, opened 2020-02-05
Store org.eclipse.jdt.core.prefs in repository

LorenzoBettini:
It is better to have org.eclipse.jdt.core.prefs in the Git repository,
with an explicit setting for the Java compiler to target 1.8 (for
compliance from the execution environment JavaSE-1.8), otherwise, if
using Java 11 for running Eclipse, the compiler will generate bytecode
for Java 11 and then you won't be able to run tests from Eclipse, with
errors of the shape

```
java.lang.UnsupportedClassVersionError:
org/pitest/pitclipse/runner/model/ModelBuilderTest has been compiled by
a more recent version of the Java Runtime (class file version 55.0),
this version of the Java Runtime only recognizes class file versions up
to 52.0
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,5 @@
 # Maven
 target/
 # Eclipse
-.settings/
 bin/
 *.versionsBackup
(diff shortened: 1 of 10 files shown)


echebbi:
Useful, thanks!

PR #98, opened 2020-02-06
don't run sonar:sonar on forked repositories

LorenzoBettini:
without this change the build fails on Travis on forked repositories, because it still tries to perform SonarQube analysis, contacting SonarCloud.
In fact the variable SONAR_SCANNER_HOME is set also when building a fork.
Use ${TRAVIS_SECURE_ENV_VARS} instead

WARNING: @echebbi you should check that on the main repository SonarCloud analysis still works :)

### DIFF
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -4,7 +4,8 @@
 # Indeed, that means that the build is performed on a branch of the
 # pitest/pitclipse repository (or an internal PR) and that Travis CI allows us to use
 # encrypted variables (e.g. Sonar's token)
-if [ "${SONAR_SCANNER_HOME}" != "" && "${TRAVIS_REPO_SLUG}" = "pitest/pitclipse" ]; then
+if [ "${SONAR_SCANNER_HOME}" != "" ] && [ "${TRAVIS_REPO_SLUG}" = "pitest/pitclipse" ]
+then
     COMMAND="mvn verify -P jacoco sonar:sonar";
 # ELSE we cannot run Sonar analysis because the branch is coming
 # from a fork and Travis CI does not allow use to use encrypted variables.
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -1,6 +1,18 @@
 #!/bin/bash
 
-COMMAND="mvn verify -P jacoco";
+# IF the variable is set, we can run Sonar analysis.
+# Indeed, that means that the build is performed on a branch of the
+# pitest/pitclipse repository and that Travis CI allows us to use
+# encrypted variables (e.g. Sonar's token)
+if [ "${TRAVIS_REPO_SLUG}" = "pitest/pitclipse" ]; then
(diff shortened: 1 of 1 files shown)


LorenzoBettini:
@echebbi from what I can see from the log it looks like this is not the right strategy for detecting a forked repo... I'll try to update the PR by using another strategy (e.g., the Git repo slug)...

LorenzoBettini:
@echebbi OK, now I'm lost... it looks like before this PR your travis build did not execute SonarCloud analysis but it was executed for my fork (?)...

would it be OK to simply skip sonar:sonar always?

LorenzoBettini:
UI tests have failed but I think that's just because they're flaky... I cannot restart the travis build myself

echebbi:
> without this change the build fails on Travis on forked repositories, because it still tries to perform SonarQube analysis, contacting SonarCloud.

Sorry about that, I would have thought I checked this before merging my changes...

> UI tests have failed but I think that's just because they're flaky... I cannot restart the travis build myself

Sadly I still didn't manage to completely fix the issue. I will merge #84 anyway: it should at least increase builds' stability.

> OK, now I'm lost... it looks like before this PR your travis build did not execute SonarCloud analysis but it was executed for my fork (?)...

What do you mean? Travis CI does trigger SonarCloud analysis, see for example the [log in latest master build](https://travis-ci.com/pitest/pitclipse/builds/147684860#L2397) (even if I have to check why the coverage is not computed anymore). 

I don't understand why Travis CI failed to build your commit [97ec935](https://github.com/pitest/pitclipse/pull/98/commits/97ec93500c280a2cda4ffb94b047c7fdd49759c0) though. Sonar was complaining about "unauthorized analysis" but only for a few files. I tried to relaunch the build just in case but it seems that your latest commit [f220e79](https://github.com/pitest/pitclipse/pull/98/commits/f220e7936ca93f1ecb08b21a825319dad239fa79) was rebuilt instead.

> would it be OK to simply skip sonar:sonar always?

SonarCloud provides important metrics, I'd prefer not to lose them.

LorenzoBettini:
The latest commit works because I disabled sonar analysis completely O:)

I now understand that sonar is not run for external PR (for security reasons) so all my experiments were wrong. I think I now understand how to fix this, I'll update the PR

LorenzoBettini:
OK, I think I got it now:

- "${SONAR_SCANNER_HOME}" != "" on external PR and on forked repositories
- "${TRAVIS_REPO_SLUG}" = "pitest/pitclipse" only on the main repository

so this should fix the problems

echebbi:
Great, looks good. I also [built an internal branch](https://travis-ci.com/pitest/pitclipse/builds/148030756) and Sonar analysis is properly triggered so I think everything works as expected.

LorenzoBettini:
@echebbi great! :)

PR #101, opened 2020-02-08
Task 99 don't store pitest jar

### LINKED ISSUE #99: Don't store pitest JARs in the Git repository
## Motivation

Currently the pitest JARs are part of the Git repository and I guess updating to a new version of pitest requires manually downloading JARs and updating the MANIFEST

## Proposed Solution

By using `maven-dependency-plugin` we can specify pitest dependencies in the POM and then have the Maven plugin create a `libs` folder inside the project (which would be excluded from Git); JARs can also be copied without their version, so that updating to a new version of pitest is just a matter of changing the version in the POM.

I'll provide a PR ASAP

LorenzoBettini:
Closes #99 

Now switching to a new version of pit is just a matter of setting a version in the POM and in the MANIFEST of org.pitest project... no need to update references to jars in other JAVA files; no need to download manually jars and most of all, jars are not stored in the git repository :)


LorenzoBettini:
@echebbi I forgot to mention that with this solution, thanks to m2e, you still have jars in Eclipse, and they are also automatically updated (if you update the version of pitest as mentioned in the PR), but they don't have to be stored in the git repo

echebbi:
Thank you very much for this: it will be of great help for future upgrades of PIT. I suspected it was somehow possible but I didn't take the time to search for a solution.

I just want to make sure that Pitclipse still works as expected after being deployed (and more specifically that PIT's plug-ins are propely found and loaded when an analysis is run) before merging this PR.

LorenzoBettini:
@echebbi I tried that also manually by generating an update site and installing pitclipse from that local site and everything still works, but of course, if you do that as well it's even better ;)

LorenzoBettini:
M2e actually automatically downloads the jars by default and no additional lifecycle mapping should be required.. I'll check this again, probably tomorrow, and in case update the documentation

LorenzoBettini:
@echebbi you're right: an additional m2e connector is required to have things automatically work (I've just tested that with a new Eclipse distribution for committers). I'll update the documentation and update the PR.

LorenzoBettini:
@echebbi I've updated the two documentation files as you requested. Please note that the procedure is a little bit more involved than the one you suggested.

IMHO, it is much better to have JARs automatically handled by Maven and not stored in the repository, especially for migrating to more recent versions of PIT.

LorenzoBettini:
Again UI tests fail because are flaky, not due to this PR :)

LorenzoBettini:
@echebbi I don't understand what the latest commit you've done "Merge branch 'master' into task_99-Don't_store_pitest_JAR" means... this is not the merge of PR, is it?

echebbi:
> this is not the merge of PR, is it?

@LorenzoBettini no, it is not. This commit is required by GitHub because your branch was not up to date with master (ideally I would just rebase your branch on master but AFAIK this [is not possible yet](https://github.com/isaacs/github/issues/1113)).

LorenzoBettini:
OK, I did not know about that otherwise I would have done that myself on my branch (as I've done for other PRs). Sorry for the inconvenience.

PR #104, opened 2020-02-15
Don't store Cucumber's JARs (fix #103)

### LINKED ISSUE #103: Don't store Cucumber JARs
## Motivation

Storing bare JARs inside the Git repository makes harder to upgrade dependencies.

## Proposed Solution

Pitest's JARs have recently been removed from the repository (see issue #99, fixed by PR #101); the same approach should work for Cucumber's JARs.

echebbi:
### Because:
 - makes easier to switch to a newer version of Cucumber (only requires to set the new version in io.cucumber's POM)
 - reduces repository's size

### Changes (based on 20c4e86):
 - remove Cucumber's JARs
 - set Cucumber as a Maven dependency of io.cucumber (in POM) so that JARs are automatically downloaded and copied in the lib/folder
 - update README and contribution guide to document M2E usage

### Why a draft?

Changes look good and the build pass as expected but JDT is not able to compile `org.pitest.pitclipse.ui.tests`: for some reason `cucumber.*` and `io.cucumber.*` imports cannot be resolved.


echebbi:
OK, so I had to explicitely state that JARs are exported (see _.classpath_). I would have thought that specifying exported packages in the MANIFEST was enough for Eclipse IDE to resolve everything but at least everything works now.

For the record, here is how I specified that JARs are exported:

image

LorenzoBettini:
The jars were exported also in the original code and the same holds for the jars of pitest

echebbi:
That's how I figured it out, I just don't understand why it was needed.

LorenzoBettini:
I guess that's because in that case there are no .class files so other projects cannot access jars unless exported (package export is just a visibility constraint, I guess)

PR #124, opened 2021-04-30
Detect when JUnit 5 is provided by Maven or Gradle

echebbi:
Contributes to #116.

This PR also introduces build qualifiers (see #120) and bumps all plug-ins and features' version to make sure users can update their installation. @LorenzoBettini would you mind reviewing the PR to make sure it looks good?

Issue #105, opened 2020-02-17
SonarCloud does not gather coverage anymore


echebbi:
## Bug description

SonarCloud always report 0% coverage

## Expected behavior

SonarCloud should report the right coverage.

## How to reproduce

Build an internal branch/PR on Travis CI.

## Additional context

It looks like Sonar team has deprecated usage of JaCoCo's `.exec` files and doesn't take them into account anymore. [Here](https://community.sonarsource.com/t/coverage-test-data-importing-jacoco-coverage-report-in-xml-format/12151) they advise to rely on XML reports instead, hopefully we can easily generate them.

LorenzoBettini:
@echebbi the jacoco-maven-plugin provides the goal `report` to do that (automatically bound to the `verify` phase): you just have to enable such a goal in the configuration of the plugin.

echebbi:
Yes, that's what is advised by the Sonar team. I gave it a try but can't make it work: either the XML report is not generated or I face errors like `Unknown block type c0` or `malformed input around byte 2`. 

Upgrading the JaCoCo plug-in to the latest version didn't help so I must be doing something wrong. I'll take a deeper look when I'll have more time.

LorenzoBettini:
I can re-enable SonarCloud analysis from GitHub Actions, but that requires using Java 11. I hope that doesn't break anything ;)

LorenzoBettini:
By the way, to do this, @echebbi I'd need you to create a secret in this repository called named `SONAR_TOKEN` with the value of your SonarCloud token (since the analysis is hosted on your SonarCloud's account).

echebbi:
@LorenzoBettini `SONAR_TOKEN` is now available on this repository.

By the way, what will be the consequences of migrating to Java 11 for the end users? I guess that they would at least need Java 11 to run the plug-in?

LorenzoBettini:
> @LorenzoBettini `SONAR_TOKEN` is now available on this repository.
> 

Thanks! I'm about to try that

> By the way, what will be the consequences of migrating to Java 11 for the end-users? I guess that they would at least need Java 11 to run the plug-in?

No consequence: the important thing is the required Java version in our bundles, which is Java 8. End-users need at least Java 8, they DON'T need Java 11. We only need to use Java 11 for building (actually only for running the SonarCloud analysis). The bytecode is generated to require Java 8. By the way, using Java 11 for building also has the benefits of switching to Tycho 2 and of building and testing against Eclipse from 2020-09 on.

PR #125, opened 2021-04-30
Task 121 switch from travis to git hub actions

LorenzoBettini:
I had many problems running SWTBot tests in GitHub Actions: the job simply hangs indefinitely and I had to cancel... I suspect there are several problems with the GTK version: we build using Oxygen that is really old and I guess it does not work with recent versions of GTK, or, at least, SWTBot cannot handle such combinations. In Travis we were using Trusty (14.04) but the oldest version provided by GitHub Actions is 16.04, which does not work either with our tests. Unfortunately, I cannot understand what's going on since after some time the job simply hangs. What I can tell for sure is that most of those SWTBot tests fail on my machine too (both from Eclipse and from the command line). We also said in the past that those tests were rather flaky...

For the moment, I had to disable those tests. I suggest we open a new issue and try to deal with those tests. I guess most of them bring no value so we can remove them.

I also cleaned up the TP a bit, by using the latest versions. I guess in the near future we should switch to a newer TP as well.

@echebbi please let me know.

P.S. Codecov and SonarCloud are not enabled yet, and I suggest we postpone that as well.


echebbi:
> I had many problems running SWTBot tests in GitHub Actions: the job simply hangs indefinitely and I had to cancel... I suspect there are several problems with the GTK version

I didn't forsee it but that was to be expected, I guess.

> What I can tell for sure is that most of those SWTBot tests fail on my machine too (both from Eclipse and from the command line). We also said in the past that those tests were rather flaky...

Yeah, you musn't touch anything during the tests and they still may fail randomly (#81). I run the tests on my machine yesterday which proves they can still pass.

> For the moment, I had to disable those tests. I suggest we open a new issue and try to deal with those tests. I guess most of them bring no value so we can remove them.

Likely the best option. Some tests are useful though because they check lots of different options and parameters but we should at least be able to turn those relevant tests into non-UI tests.

> I also cleaned up the TP a bit, by using the latest versions. I guess in the near future we should switch to a newer TP as well.

I'm completely open to such a switch, we just have to discuss what is the oldest version of Eclipse IDE we want to support.

> Codecov and SonarCloud are not enabled yet, and I suggest we postpone that as well.

They used to work but SonarCloud broke a while ago and I never took the time to fix it (#105). Now they don't even support Java 8 anymore, causing the whole pipeline to fail. In the other PR I [disabled it](https://github.com/pitest/pitclipse/pull/124/commits/8485939fbbf5097b02d196d4ac183dd07d200209) until we can fix it.

LorenzoBettini:
@echebbi so if it's OK I'd merge it

LorenzoBettini:
@echebbi I created and pushed a Git tag for the new release, but on GitHub there's still the release 2.1.0 tagged as "Latest" and I don't know how to fix that...

echebbi:
@LorenzoBettini Releases have to be created manually. First go to https://github.com/pitest/pitclipse/releases (accessible from Pitclipse's main page > _Releases_) then click on _Draft a new release_. For future tags we may want to automate thanks to the [GitHub Create Tag Release](https://github.com/marketplace/actions/github-create-tag-release) Action.

LorenzoBettini:
OK, for the moment I've promoted the release created with the tag as the latest one.

By the way, currently releasing a new update site from GitHub Actions is still not enabled.
For the moment, we administrators can publish a new update site from our computer:

```
mvn deploy -Prelease-composite
```

or, to publish a new snapshot:

```
mvn deploy -Prelease-composite,release-snapshots
```

During the git clone/push you might be asked for your password or SSH passphrase unless you have some keyring mechanism on your computer.

To publish from GitHub Actions we need a personal access token in the GitHub `pitest` organization. Or maybe it might be enough to create one for our own user and make it available at the organization level.

PR #126, opened 2021-05-06
120 deploy to GitHub pages

### LINKED ISSUE #120: Pitclipse JUnit 5 support cannot be (easily) installed
## Bug description

This is a follow-up of #116 but I decided to create a new issue to draw attention. I think this should be marked as **critical**.

The current version 2.1.1 deployed to the composite and main update site https://dl.bintray.com/kazejiyu/Pitclipse/updates/ will not allow you to properly and easily install JUnit 5 support of pitclipse (that is, the bundle `org.pitest.pitest-junit5-plugin`). 

That is due to the fact that the feature `org.pitest.feature` has been published several times in different children of the main composite site with the same version 1.4.11 but with different contents (i.e., requirements of `org.pitest.pitest-junit5-plugin`). That is, in the child update site http://dl.bintray.com/kazejiyu/Pitclipse/releases/2.1.0/ the feature does NOT include `org.pitest.pitest-junit5-plugin`, while in the child update site http://dl.bintray.com/kazejiyu/Pitclipse/releases/2.1.1/ the feature INCLUDES `org.pitest.pitest-junit5-plugin`.

However, Eclipse p2 does not check the contents when installing and takes into consideration the version only. If it inspects the first update site child first (which I think it's the default behavior, but I cannot guarantee that this strategy is deterministic), it will install `org.pitest.feature` 1.4.11 WITHOUT `org.pitest.pitest-junit5-plugin`. After that, there's no way of forcing it to install (actually update) `org.pitest.feature` 1.4.11 even if specifying only the http://dl.bintray.com/kazejiyu/Pitclipse/releases/2.1.1/. In fact, that feature has already been installed, and, even there's a new JAR with different contents it will be ignored since it has exactly the same version (again, actually p2 will not even detect that there's a new version since the versions are exactly the same).

Thus, after the damage is done, fixing it might not be easy for standard users: you have to uninstall "Pitclipse UI", remove the composite update site (https://dl.bintray.com/kazejiyu/Pitclipse/updates/) add only the latest single update site (http://dl.bintray.com/kazejiyu/Pitclipse/releases/2.1.1/) and finally install "Pitclipse UI" that will also pick `org.pitest.pitest-junit5-plugin`).

In the presence of p2 caches, the fix might also not even work.

One of the golden rules of p2 is: NEVER EVER publish p2 artifacts with the same version but with different contents: you will destroy completely the user experience and it might not even be easy for non-advanced users to fix the problem. In general, it is best practice to always publish p2 artifacts with a build qualifier to make them always different. Maven/Tycho automatically handles that if you add .qualifier and -SNAPSHOT to Eclipse metadata and POMs, respectively: it will replace the qualifier with a build qualifier using the current time stamp. It's also possible to have Tycho produce the same build qualifier if nothing has changed (using Git information), but again this requires careful attention. This way, even if you publish exactly the same artifacts (features and bundles) with exactly the same contents they will have different versions in the end.

Please let me know if you want me to open a new issue for that and if you need help with a PR.

Of course, the alternative is to make sure to bump at least the "patch" versions of all the Pitclipse bundles and features and publish a new child update site in the composite main site. However, as said above, not using build qualifiers in version is error prone.

## Expected behavior

By installing "Pitclipse UI", the JUnit 5 support should be also installed.

## How to reproduce

1. Download a fresh Java distribution from Eclipse
2. "Install New Software..." and "Work with" https://dl.bintray.com/kazejiyu/Pitclipse/updates/
3. Select "Pitclipse UI" version 2.1.1 (currently the latest one)
4. Press "Finish" and restart Eclipse after installation
5. Inspect the "Installation Details" (using "About Eclipse IDE) and in the tab "Plug-ins" verify that `org.pitest.pitest-junit5-plugin` is NOT installed.

### LINKED ISSUE #122: Switch from Bintray to GitHub Pages for serving the Eclipse update site
## Motivation

Bintray is shutting down, so we cannot serve the Eclipse update site from there...

## Proposed Solution

I suggest we switch to GitHub Pages for the Eclipse (composite) update site. I already have a working example for that. This requires creating a new GitHub repository for hosting the update site (@hcoles, @echebbi I think I can create that myself in the organization with my admin rights). While hosting binaries on GitHub pages is not ideal, I don't foresee a lot of traffic so we should be fine (and in my experiments, it's pretty fast).

By the way, Bintray is shutting down really soon (I guess the end of March), so we should act pretty quickly.

LorenzoBettini:
Closes #120
Closes #122


LorenzoBettini:
@echebbi I've concluded this task and I have already deployed both a snapshot in the snapshot update site and a release in the release update site.
Everything seems to work (tested manually by installing from the update sites in an Eclipse distribution).
I'm going to update the README and the Marketplace.
I'd also merge this PR, OK?

echebbi:
> @echebbi I've concluded this task and I have already deployed both a snapshot in the snapshot update site and a release in the release update site.
Everything seems to work (tested manually by installing from the update sites in an Eclipse distribution).
I'm going to update the README and the Marketplace.

Wonderful, thank you very much for you work!

> I'd also merge this PR, OK?

That looks good to me 👍

PR #132, opened 2021-05-07
Refactoring

LorenzoBettini:
PitclipseService was used only in its own PitclipseServiceTest so I removed them both


LorenzoBettini:
@echebbi what do you think?

LorenzoBettini:
@echebbi I'd like to merge it is that OK?

echebbi:
@LorenzoBettini if the files aren't actually used, why did you moved them instead of merely removed them?

LorenzoBettini:
> @LorenzoBettini if the files aren't actually used, why did you moved them instead of merely removed them?

I just wanted to be sure ;)
if you prefer I'll remove them directly

echebbi:
Yes, I think it's cleaner that way. If those files ever end up appearing useful we'll still be able to recover them with Git.

PR #133, opened 2021-05-07
Sonar issues

LorenzoBettini:
This should fix most blocker issues in SonarCloud (they were false positives)


LorenzoBettini:
@echebbi I'd like to merge it is that OK?

echebbi (approved):

echebbi:
Yes, looks good to me!

PR #140, opened 2021-07-10
Task 85 speed up UI tests

### LINKED ISSUE #85: Speed up UI tests
## Motivation

UI tests are really slow and run for an average of 8 minutes. I believe that this time can be improved.

## Implementation

Currently, during UI tests a lot of projects, classes and methods are created. The process relies on SWTBot and mimic user's actions by, for instance, using wizards. This takes a lot of time.

We should be able to optimize this, for instance by creating the projects in advance and then importing them dynamically before running tests. This would also simplify a lot of code and would likely make easier to check and update the projects used during tests.

LorenzoBettini:
Closes #85 

This aims at speeding up UI tests by NOT creating projects programmatically: it imports already defined projects for testing purposes and avoids recreating over and over again.


LorenzoBettini:
@echebbi as usual, if you have time to have a look at that, otherwise I'll merge this in the next few days

Issue #138, opened 2021-07-09
Maven requirements aren't correct


JKutscha:
## Bug description
The requirements lists Maven 3.x, but you can't use it with Maven 3.0.5.

## How to reproduce

1. Install Maven 3.0.5
2. Run mvn or mvn clean verify in the main folder 

## Fix
Just update the README requirements to Maven 3.1.x or higher perhaps?

LorenzoBettini:
@JKutscha could you please be more specific on the error you get?
In any case, for sure Maven 3.0.5 is rather old. I think at least Maven 3.6.3 is required.
Of course, if you don't mind, please feel free to create a PR for the README :)

JKutscha:
The error is the following:

/home/pitclipse[master]$ mvn -e                                                                                                                                                                                                                                    
[INFO] Error stacktraces are turned on.                                                                                                                                                                                                                                                   
[INFO] Scanning for projects...                                                                                                                                                                                                                                                           
[ERROR] The build could not read 2 projects -> [Help 1]                                                                                                                                                                                                                                   
org.apache.maven.project.ProjectBuildingException: Some problems were encountered while processing the POMs:                                                                                                                                                                              
[ERROR] Child module /home/pitclipse/bundles/org.pitest.pitclipse.core/pom.xml of /home/pitclipse/bundles/pom.xml does not exist @                                                                                                                              
[ERROR] Child module /home/pitclipse/bundles/org.pitest.pitclipse.launch/pom.xml of /home/pitclipse/bundles/pom.xml does not exist @                                                                                                                            
[ERROR] Child module /home/pitclipse/bundles/org.pitest.pitclipse.launch.ui/pom.xml of /home/pitclipse/bundles/pom.xml does not exist @                                                                                                                         
[ERROR] Child module /home/pitclipse/bundles/org.pitest.pitclipse.listeners/pom.xml of /home/pitclipse/bundles/pom.xml does not exist @                                                                                                                         
[ERROR] Child module /home/pitclipse/bundles/org.pitest.pitclipse.preferences.ui/pom.xml of /home/pitclipse/bundles/pom.xml does not exist @                                                                                                                    
[ERROR] Child module /home/pitclipse/bundles/org.pitest.pitclipse.runner/pom.xml of /home/pitclipse/bundles/pom.xml does not exist @                                                                                                                            
[...]

PR #141, opened 2021-07-12
Changed requirements of mavevn to 3.6.3 (#138)

JKutscha:
Earlier versions of maven don't work properly


LorenzoBettini:
Closes #138

Issue #62, opened 2015-11-17
Can only navigate forward in the coverage report window


grahamtt:
A 'back' button would be very helpful

AFulgens:
Context menu (right click) has a "Back" option. It's suboptimal, but works :)

echebbi:
Fixed by #150

Issue #114, opened 2020-03-01
Provide a finer grain control on mutators to enable


echebbi:
## Motivation

As suggested by @LorenzoBettini in #77 ([here](https://github.com/pitest/pitclipse/issues/77#issuecomment-584073984)) 

## Proposed Solution

Enhance the preferences page with a table displaying the full list of options (we should be able to support [all the options offered by Pitest](http://pitest.org/quickstart/commandline/)) which a checkbox to enable/disable each.

JKutscha:
I would like to work on that topic, but would like to add this in the run config to be more flexible.

LorenzoBettini:
That's great! But please, keep in mind that you should also write a few UI tests for such new features.
Until https://github.com/pitest/pitclipse/pull/140 is merged to master, you should start from the corresponding branch. You can take inspiration from the tests in the project `org.pitest.pitclipse.ui.tests` and the package `org.pitest.pitclipse.ui.tests`.

JKutscha:
I have looked into it, I just realized there are 150 mutators. Would you think to have full control over them would be nice or should we just present some of them?

echebbi:
Hey there! Just so you know I've already worked on this feature in the [`114-provide-finer-grain-control-over-mutators`](https://github.com/pitest/pitclipse/tree/114-provide-finer-grain-control-over-mutators) branch. It's a bit old (about 1 year old) so I don't know if it can be rebased on master easily but you should at least be able to reuse some code.

The branch adds a table to the Pit tab in the Run Configuration view that allows to select the mutators to run (the user can either selects pre-defined groups or selects a custom subset of mutators):

![image](https://user-images.githubusercontent.com/25926433/126321187-7def7090-10b5-42c9-8530-16c693733368.png)

I think I added all the mutators that were described in [Pitest's website](https://pitest.org/quickstart/mutators/) at that time; they are all defined in the [Mutators enum](https://github.com/pitest/pitclipse/blob/12683d94432edeb4fbdeacd3fff42cb5a7e82b6b/bundles/org.pitest.pitclipse.core/src/org/pitest/pitclipse/core/Mutators.java#L6-L34).

I think that I was almost done with the UI part but didn't start to actually use it to configure the Pitest run. So the `performApply` is likely not implemented in the Mutators tab to persist the preferences and the Pitest runner doesn't get the selected mutators either. 

[...]

PR #148, opened 2021-07-13
PIT Summary view enhancements

JKutscha:
* view tries to open last active editor file
* added navigation buttons:
    - home: get back to index.html in basedir
    - <: back one page
    - \>: forward one page


JKutscha:
To fix issues #62 and improve usability.

LorenzoBettini:
Thanks for the contribution, but please make sure you configure your eclipse to use spaces instead of tabs, and then resubmit

JKutscha:
I have run all tests locally and they pass. I'm not sure why they're failing here.

LorenzoBettini:
> I have run all tests locally and they pass. I'm not sure why they're failing here.

The one on Windows seems to have failed because it couldn't download a few Maven dependencies.. that happens sometimes... I restarted the build

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -16,28 +16,73 @@
...
+                @Override
+                public void run() {
+                    if (isBrowserSet()) {
+                        browser.forward();
+                    }
+                }
+            };
+            IActionBars actionBars = getViewSite().getActionBars();
+            IToolBarManager toolBar = actionBars.getToolBarManager();
+            toolBar.add(backButton);
```
LorenzoBettini, line 82: If I understand correctly (I haven't tried that locally), the buttons are added to the Eclipse toolbar, not the the PIT view?

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -53,9 +98,57 @@ public void setFocus() {
     @Override
     public synchronized void update(File result) {
         if (result == null) {
+            homeUrlString = null;
             browser.setText("<html/>");
         } else {
-            browser.setUrl(result.toURI().toString());
+            homeUrlString = result.toURI().toString();
+            // fallback, if the file does not exist
```
LorenzoBettini, line 105: This comment refers to the result file or what? There doesn't seem to be any fallback here

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -53,9 +98,57 @@ public void setFocus() {
...
+     * Gets the path of the given path in context with the base dir of an result
+     * 
+     * @param filePath of file, which should be present in pit result
+     * @param baseDir  of the results in HTML format
+     * @return file path in context of the base dir
+     */
+    public static String getFilePathInResultDir(String filePath, String baseDir) {
+        // replace separator chars with dots for regEx
+        filePath = filePath.replace(File.separatorChar, '.');
+        Pattern pattern = Pattern.compile("(.*\\.src\\.)(.*)(\\.)(.*\\.java)");
```
LorenzoBettini, line 142: You cannot assume the Java files are always in a folder called `src`

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -53,9 +98,57 @@ public void setFocus() {
...
+        Matcher matcher = pattern.matcher(filePath);
+        if (matcher.find() && matcher.groupCount() == 4) {
+            // group 1 and 3 are not from interest
+            return baseDir + File.separatorChar + matcher.group(2) + File.separatorChar + matcher.group(4) + ".html";
+        }
+        return null;
+    }
+
+    private boolean isBrowserSet() {
+        return browser != null && homeUrlString != null;
```
LorenzoBettini, line 152: how can the browser be null?

LorenzoBettini (changes requested):
Besides the single comments, I'd like to see a few UI tests to cover also corner-case situations :)
With that respect, please have a look at this comment https://github.com/pitest/pitclipse/issues/114#issuecomment-879681661

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
JKutscha: It is added to the view itself:
grafik

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
JKutscha: The fallback is in line 106. The normal index.html is opened and only a new url is set, if a new url points to an existing file.

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
JKutscha: If in createPartControl() an exception is thrown while creating the browser.

JKutscha:
> 
> 
> Besides the single comments, I'd like to see a few UI tests to cover also corner-case situations :)
> With that respect, please have a look at this comment [#114 (comment)](https://github.com/pitest/pitclipse/issues/114#issuecomment-879681661)

Should I wait with committing the UI tests until #140 is merged or should I add them to this pr?

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
LorenzoBettini: OK!

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
LorenzoBettini: I'd avoid conditions that cannot be easily tested and in any case, setting the URL even if it's null or if the file does not exist should be safe.
Otherwise, if all the conditions can be covered by test then it's fine to me

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -53,9 +98,61 @@ public void setFocus() {
...
+     * Looks up the last active file from an editor and constructs a path to its
+     * page of the result
+     * 
+     * @param result file which is used for the path beginning
+     * @return String which represents the path in the result folder or null, if the
+     *         file form the editor is no .java file or no FileEditor was active
+     *         last
+     */
+    private String getUrlToLastActiveClass(File result) {
+        IEditorInput input = getSite().getPage().getActiveEditor().getEditorInput();
```
LorenzoBettini, line 124: What if there's no active editor? NPE?

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -53,9 +98,61 @@ public void setFocus() {
...
+        if (input instanceof FileEditorInput) {
+            IFile file = ((FileEditorInput) input).getFile();
```
LorenzoBettini, line 126: when can an input be different from a FileEditorInput?

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
LorenzoBettini: Well in that case we have more serious problems anyway, so it's useless to check for the browser to be null ;)
even if the homeUrlString is null it should be safe to go back or forward in the browser, shouldn't it? So this check might be useless

LorenzoBettini (changes requested):

LorenzoBettini:
> > Besides the single comments, I'd like to see a few UI tests to cover also corner-case situations :)
> > With that respect, please have a look at this comment [#114 (comment)](https://github.com/pitest/pitclipse/issues/114#issuecomment-879681661)
> 
> Should I wait with committing the UI tests until #140 is merged or should I add them to this pr?

It's been just merged :)

you could merge or rebase your branch

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
JKutscha: Should all be tested and I agree that it would be safe to set the url, but the user would get an "no such file" message in the browser.

LorenzoBettini:
@JKutscha I'm not sure I understand what "view tries to open last active editor file"... it looks like that part is not easy to handle especially because it's based on HTML contents created into several directories, whose layout might also change in the future (if PIT changes)... maybe a first PR with just navigation buttons would be useful

JKutscha:
> 
> 
> @JKutscha I'm not sure I understand what "view tries to open last active editor file"... it looks like that part is not easy to handle especially because it's based on HTML contents created into several directories, whose layout might also change in the future (if PIT changes)... maybe a first PR with just navigation buttons would be useful

Would you explain what part you don't understand, that I can explain it to you.

LorenzoBettini:
> "view tries to open last active editor file"

I don't understand what the above part should do...

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitUiUpdatePublisher.java
```diff
@@ -44,9 +51,33 @@ public void changed(ProgressEvent event) {
...
+     * Get the index.html of the results, even if the browser shows a different file
+     * of the results
+     * @return html text of the index.html as a String
+     */
+    private String getHtml() {
+        // assume url is path to index.html and top folder should represent date and
+        // time
+        // remove os differences in path
+        final Pattern pattern;
+        if (WINDOWS) {
```
LorenzoBettini, line 68: This will be impossible to cover since we do code coverage only in Linux.
Moreover, I'd avoid solutions that are based on the OS.
That's also why I suggested to focus on the browser navigation buttons and postpone this other feature to the future (though, as I said, I doubt it can be implemented safely since it highly relies on the format of the HTML that might change in the future versions of PIT, making it a problem to maintain).

In any case, if you want to implement such a mechanism, I'd strongly suggest you factor out this code into a separate class that does not need the UI and can be tested in isolation, e.g., in `runner.tests`. For example, you can put into the `runner.tests` a directory with some contents and run tests against such contents.

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
LorenzoBettini: Well, if that's because there's no such file then the user should get that error, or I don't understand the scenario...

JKutscha:
> I don't understand what the above part should do...

@LorenzoBettini if you start a pitclipse run and you opened Foo.java last. Then the PIT Summary View would show you the results of the Foo.java, if it exists in the mutated files.

You could see it like the "Link with Editor" between the package explorer and the editor, but it influences the view only on start up. Is it more clear now?

The feature itself works, but the problem I'm having is, that the run data, like coverage or mutated classes, is parsed from the open HTML page in the pit summary view and used to prove if the test were run correct.

Would a button that jumps to the corresponding html page of the opened class be more applicable?

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -53,9 +96,57 @@ public void setFocus() {
...
+     * Looks up the last active file from an editor and constructs a path to its
+     * page of the result
+     * 
+     * @param result file which is used for the path beginning
+     * @return String which represents the path in the result folder or null, if the
+     *         file form the editor is no .java file or no FileEditor was active
+     *         last
+     */
+    private String getUrlToLastActiveClass(File result) {
+        IEditorPart editor = getSite().getPage().getActiveEditor();
```
LorenzoBettini, line 122: I'm wondering if you could do something along the lines of `org.pitest.pitclipse.launch.ui.PitLaunchShortcut.getLaunchableResource(IEditorPart)`, which also relies on `org.eclipse.jdt.ui.JavaUI.getEditorInputTypeRoot(IEditorInput)`

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PackageExplorer.java
```diff
@@ -136,7 +136,7 @@ public void openClass(ClassContext context) {
     }
 
     public void selectPackage(PackageContext context) {
-        getPackage(context).select();
+        getPackage(context).select().expand();
```
LorenzoBettini, line 139: mh... this should only select the package, maybe you need another method that also expands that?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
```diff
@@ -46,4 +61,67 @@ public double getOverallCoverage() {
...
+    }
+
+    public void clickButtonWithText(String text) {
+        showViewIfNotOpen();
+        toolbarButton(text).click();
+    }
+
+    /**
+     * Helper method, because other method works with tooltip. Gets the toolbar
+     * button matching the given toolbar button.
```
LorenzoBettini, line 79: Maybe this means that we should add tooltips on those buttons? Would that be useful?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitSummaryViewTest.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * @author Jonas Kutscha
+ * 
+ */
+@RunWith(SWTBotJunit4ClassRunner.class)
+public class PitclipsePitSummaryViewTest extends AbstractPitclipseSWTBotTest {
+    private static final String TEST_PROJECT = "org.pitest.pitclipse.testprojects.twoclasses";
+    private static final String FOO_BAR_PACKAGE = "foo.bar";
+    private static final String FOO_CLASS = "Foo";
+    private static final String TEST_PROJECT_WITH_DIFFERENT_SRC = "org.pitest.pitclipse.testprojects.differentSrc";
```
LorenzoBettini, line 28: Please don't use capital letters in project names (nor in the project directory).
Please make sure you change consistently the name and the directory to `org.pitest.pitclipse.testprojects.differentsrc`

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitSummaryViewTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Test
+    public void navigateWithButtons() throws CoreException {
+        final long timeout = SWTBotPreferences.TIMEOUT;
+        openEditor(FOO_CLASS, FOO_PACKAGE, TEST_PROJECT_WITH_DIFFERENT_SRC);
+        runPackageTest(FOO_PACKAGE, TEST_PROJECT_WITH_DIFFERENT_SRC);
+        coverageReportGenerated(1, 80, 0);
+        // set timeout to 5 ms, because the browser should not load pages and run into
+        // timeout
+        SWTBotPreferences.TIMEOUT = 5;
```
LorenzoBettini, line 69: If you change such a value, you must make sure you always reset it, even if the test fails (or this will break other subsequent tests): either you use a try...finally block (and you reset it in the finally block) or you use `@Before` and `@After`

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/util/ViewUtil.java
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * Opens the view in the given node and returns its handle, identified with the id.
+     * @param bot      used for the menus
+     * @param viewName name of the view seen, in "Show View" shell
+     * @param nodeName name of the node, where the view can be found, in "Show View" shell
+     * @param viewId   the view id which identifies the view
+     * @return the handle of the view
+     */
+    public static SWTBotView openViewById(SWTWorkbenchBot bot, String viewName, String nodeName, String viewId) {
+        bot.menu("Window").menu("Show View").menu("Other...").click();
```
LorenzoBettini, line 22: If you know the view id, it's better to open a view programmatically, instead of mimicking the operations relying on Eclipse dialogs (in SWTBot tests this might easily become flaky).

For example something like

```java
PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage().showView("viewId");
```

you have to do that inside a UI thread block in SWTBot, similar to what is done in `org.pitest.pitclipse.ui.tests.AbstractPitclipseSWTBotTest.openJavaPerspective()`

then, outside the block, you can get hold on the opened view with SWTBot: `bot.viewById(viewId)`

tests/testprojects/org.pitest.pitclipse.testprojects.differentSrc/.project
```diff
@@ -0,0 +1,28 @@
...
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.jdt.core.javanature</nature>
+	</natures>
+	<filteredResources>
+		<filter>
+			<id>1626263920267</id>
```
LorenzoBettini, line 19: This filter looks like bogus to me...

tests/testprojects/org.pitest.pitclipse.testprojects.differentSrc/README
```diff
@@ -0,0 +1,2 @@
+This is a project imported by our SWTBot tests (org.pitest.pitclipse.ui.tests)
+in particular, by the PitclipseOptionsTest and PitclipsePitMutationsViewTest.
```
LorenzoBettini, line 2: Please update the line to reflect the actual test where it's being used

LorenzoBettini (changes requested):
Here are a few comments.

I still think that this PR should focus only on the navigation buttons ;)

JKutscha:
@LorenzoBettini I will delete this pr and open a new one solely on the navigation buttons.
And will try to implement the other feature in another way, which hopefully is more clear.

Thanks for your patience.

LorenzoBettini:
> @LorenzoBettini I will delete this pr and open a new one solely on the navigation buttons.
> And will try to implement the other feature in another way, which hopefully is more clear.
> 
> Thanks for your patience.

Thank you for your contributions! :)

PR #150, opened 2021-07-16
Summary view navigation

JKutscha:
Issue #62 
* Added navigation via Buttons in Summary view


JKutscha:
@LorenzoBettini hi again,

This should all run fine. I run all tests with my eclipse on Windows and Linux successfully, but for whatever reason on Windows mvn clean verify fails. Can you help me and tell me how this can be?

LorenzoBettini:
You removed the PitNotifier and this breaks the synchronization I guess and makes the tests unreliable.

JKutscha:
> You removed the PitNotifier and this breaks the synchronization I guess and makes the tests unreliable.

@LorenzoBettini now they are reliable again in regard of waiting for the result.

JKutscha:
@LorenzoBettini I ran the test now multiple times on my Windows PC and they are always successful. Would you mind taking a look?

LorenzoBettini:
You removed the PitNotifier.java and you removed a few checks in the existing tests...
Please don't change the architecture of the existing testing infrastructure...
if you can't succeed in testing the new scenarios I'll try to write the tests myself on top of your new feature, but please first restore the existing working tests

JKutscha:
I only removed the part of the test, where empty results are expected and no html is created and the checks were to look at the html.

Furthermore all checks which were present are working fine and only my test case is failing, but only online. Tested them on windows and linux multiple times.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitNotifier.java
```diff
@@ -1,43 +0,0 @@
-/*******************************************************************************
```
LorenzoBettini, line 1: This file has been removed!

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipseUiRunnerTest.java
```diff
@@ -34,7 +34,6 @@ public void emptyClassAndEmptyTest() throws CoreException {
         runTest(FOO_TEST_CLASS, FOO_BAR_PACKAGE, TEST_PROJECT);
         consoleContains(0, 0, 100, 0, 0);
         mutationsAre(Collections.emptyList());
-        coverageReportGenerated(0, 100, 100);
```
LorenzoBettini, line 37: this check has been removed

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipseUiRunnerTest.java
```diff
@@ -48,7 +47,6 @@ public void emptyClassAndEmptyTestMethod() throws CoreException {
...
```
LorenzoBettini, line 51: this check has been removed

LorenzoBettini (changes requested):

LorenzoBettini:
> I only removed the part of the test, where empty results are expected and no html is created and the checks were to look at the html.
> 
> Furthermore all checks which were present are working fine and only my test case is failing, but only online. Tested them on windows and linux multiple times.

See my detailed comments

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitNotifier.java
JKutscha: Is now using an independent extension point with the class PitResultsNotifier.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipseUiRunnerTest.java
JKutscha: This check internally refers to the created HTML page from Pit, which are not created anymore since the new update, if nothing is mutated.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipseUiRunnerTest.java
JKutscha: See comment above.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitNotifier.java
LorenzoBettini: But the extension point is still there and it looks like the new mechanism is not working anyway... why do you need to introduce a new mechanism?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipseUiRunnerTest.java
LorenzoBettini: But Pitclipse will create an empty HTML will it not?

LorenzoBettini:
> Furthermore all checks which were present are working fine and only my test case is failing, but only online. Tested them on windows and linux multiple times.

but did you also run the Maven build?

The failure you get in the CI is as follows:

```
Failures: 
  PitclipsePitSummaryViewTest.navigateWithButtons:60 
Expected: a string ending with "index.html"
     but: was "file:///D:/a/pitclipse/pitclipse/tests/org.pitest.pitclipse.ui.tests/target/work/data/.metadata/.plugins/org.pitest.pitclipse.core/html_results/202107180752/foo.bar/Foo.java.html"
```

JKutscha:
Yes, I also ran maven.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitNotifier.java
JKutscha: Which extension point do you mean?
The fragment.xml only has the new extension point.

I don't see why the new mechanism is not working. What is not working?
I reused the mechanism which is used to supply the Pit Summary view directly and removed the other dependency because  this one would function as this:

1. pit started
2. pit done and notify pit summary view with the html files location.
3. notify that pit summary view has results
4. grab html text from browser of summary view and parse this
5. use parsed html text for proof of tests

New implementation of mine:
1. pit started
2. pit done notify summary view and PitResultsNotifier
3. Parse html file itself and present it for tests

So I deleted the dependency to the pit summary view, which is not necessary.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipseUiRunnerTest.java
JKutscha: No, the file is not created or at least not communicated with this extension point.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitNotifier.java
JKutscha: > But the extension point is still there and it looks like the new mechanism is not working anyway... why do you need to introduce a new mechanism?

If you meant the class with the extension point "PitResultsView", I'm sorry I forgot to delete it. It is now gone in the newest commit, because it is not used.

JKutscha:
> 
> 
> Yes, I also ran maven.

grafik

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitNotifier.java
LorenzoBettini: This class was crucial for synchronization if I remember correctly and it was also properly reset before running a new PIT run

LorenzoBettini:
Usually when it works locally but not on the CI you have synchronization problems, maybe because the CI virtual environment is faster than your machine or vice-versa. Note that now you have a failure in the macOS CI environment, while before you had it in the Windows virtual environment.

Once again, my suggestion is to keep the implementation extremely simple (e.g., don't keep track of the url string, just set it in the browser and let the buttons do the rest); similarly, tests must also be simple and reproducible. For example, don't rely on the browser to be reactive on changing pages. If I remember correctly there are lots of asynchronous operations going on there. Moreover, rely on the summary view automatically opened after running PIT, and don't try to open the view yourself: the browser might be out of date.

Once again, if you want I can simplify both the implementation and try to implement the tests.

JKutscha:
> 
> 
> Usually when it works locally but not on the CI you have synchronization problems, maybe because the CI virtual environment is faster than your machine or vice-versa. Note that now you have a failure in the macOS CI environment, while before you had it in the Windows virtual environment.
> 
> Once again, my suggestion is to keep the implementation extremely simple (e.g., don't keep track of the url string, just set it in the browser and let the buttons do the rest); similarly, tests must also be simple and reproducible. For example, don't rely on the browser to be reactive on changing pages. If I remember correctly there are lots of asynchronous operations going on there. Moreover, rely on the summary view automatically opened after running PIT, and don't try to open the view yourself: the browser might be out of date.
> 
> Once again, if you want I can simplify both the implementation and try to implement the tests.

What do you mean by don't keep track of the url string? Do you refer to the "Home" button?

How would you test the navigation with the buttons, if you dont wait for the pages to load?

The summary view itself gets disposed and the browser can't be out of date, because a new one gets created while opening.

If you want to change things feel free, but I want to know why things are changed. So I can learn from that for the future.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitNotifier.java
JKutscha: The synchronization is now done with PitResultsNotifier and works well as far as  the tests showed.
It is also reset in the same place as the old one. The behavior is almost the same as before, but without relying on the Pit Summary View.

JKutscha:
It seems the issue we were having with macOS were causing the reliability issues I was having.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitNotifier.java
JKutscha: @LorenzoBettini can I assume this is resolved?

JKutscha:
> 
> It seems the issue we were having with macOS were causing the reliability issues I was having.

Or not ...

@LorenzoBettini Is there any way to know which version of Eclipse and which Browser is used on the server site?

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -29,13 +32,54 @@
  * A view displaying the HTML report generated by PIT.
  */
 public class PitView extends ViewPart implements SummaryView {
+    public final static String VIEW_ID = "org.pitest.pitclipse.ui.view.PitView";
```
echebbi, line 35: `static` should be written before `final`: `public static final`

This is a naming convention. You can install the Checkstyle plug-in to catch such warnings.

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -29,13 +32,54 @@
...
     private Browser browser = null;
+    private PitUiUpdatePublisher publisher = null;
+    private String homeUrlString = null;
+    private Action homeAction;
+    public static String HOME_BUTTON_TEXT = "Home";
```
echebbi, line 40: Make it final (and move the constant to the top of the class). Also does it really have to be public?

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -29,13 +32,54 @@
...
+    private Action backButton;
+    public static String BACK_BUTTON_TEXT = "<";
```
echebbi, line 42: Make it final (and move the constant to the top of the class). Also does it really have to be public?

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -29,13 +32,54 @@
...
+    private Action forwardButton;
+    public static String FORWARD_BUTTON_TEXT = ">";
```
echebbi, line 44: Make it final (and move the constant to the top of the class). Also does it really have to be public?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
```diff
@@ -16,34 +16,86 @@
...
     }
 
-    public void waitForUpdate() {
-        try {
-            lastResults = PitNotifier.INSTANCE.getResults();
-        } catch (InterruptedException e) {
-            throw new StepException(e);
+    public void showViewIfNotOpen() {
+        if (summaryView == null) {
+            summaryView = bot.viewByTitle(SUMMARY_VIEW_TITLE);
```
echebbi, line 38: Do we really need a bot? I think we could get a reference to `PitView` instead.

Moreover, I think that the methods `goBack()`, `goForward()`, `goHome()` and `getUrl()` could be implemented directly in the `PitView` class.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
```diff
@@ -16,34 +16,86 @@
...
+        showViewIfNotOpen();
+        // use this instead of the SWTBrowser.execute(), because we need syncExec
+        final String link = "window.location.href = '" + hyperLinkText + "'";
+        summaryView.bot().browser().waitForPageLoaded();
+        UIThreadRunnable.syncExec(new VoidResult() {
+            @Override
+            public void run() {
+                summaryView.bot().browser().widget.execute(link);
+            }
+        });
```
echebbi, line 87: Use a lambda expression instead of an anonymous class when possible:
```java
UIThreadRunnable.syncExec(() ->
        summaryView.bot().browser().widget.execute(link)
);
```

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/steps/PitclipseSteps.java
```diff
@@ -60,24 +59,24 @@
...
+            throws CoreException {
         // No need to do a full build: we should be now synchronized with building
         // Build the whole workspace to prevent random compilation failures
-        // ResourcesPlugin.getWorkspace().build(IncrementalProjectBuilder.FULL_BUILD, new NullProgressMonitor());
-        System.out.println
-            (String.format("Run PIT on: %s %s.%s",
-                projectName, packageName, testClassName));
+        // ResourcesPlugin.getWorkspace().build(IncrementalProjectBuilder.FULL_BUILD,
+        // new NullProgressMonitor());
+        System.out.println(String.format("Run PIT on: %s %s.%s", projectName, packageName, testClassName));
```
echebbi, line 68: Avoid reformatting existing code when the PR is not about reformatting: it can be confusing and adds unnecessary review work.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitResultNotifier.java
```diff
@@ -16,36 +16,97 @@
...
 
-    private void notifiyTestsOfHtmlResults(PitUiUpdate updateEvent) {
-        PitResultsView view = buildResultsView(updateEvent);
-        tryNotifyResults(view);
-    }
+        private Summary summary;
+        private SWTWorkbenchBot bot;
+
+        public int getClasses() {
+            bot.waitUntil(new ICondition() {
```
echebbi, line 37: Replace `ICondition` with
`org.eclipse.swtbot.swt.finder.waits.DefaultCondition`. You won't have to implement `init` anymore.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitResultNotifier.java
```diff
@@ -16,36 +16,97 @@
...
+
+                public String getFailureMessage() {
+                    return "No summary set.";
+                }
+            }, SWTBotPreferences.TIMEOUT);
+            return summary.getClasses();
+        }
+
+        public double getCodeCoverage() {
+            bot.waitUntil(new ICondition() {
```
echebbi, line 54: Replace `ICondition` with
`org.eclipse.swtbot.swt.finder.waits.DefaultCondition`. You won't have to implement `init` anymore.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitResultNotifier.java
```diff
@@ -16,36 +16,97 @@
...
 
-    private PitResultsView buildResultsView(PitUiUpdate results) {
-        ResultsParser parser = new ResultsParser(results.getHtml());
-        Summary summary = parser.getSummary();
-        PitResultsView view = PitResultsView.builder().withClassesTested(summary.getClasses())
-                .withTotalCoverage(summary.getCodeCoverage()).withMutationCoverage(summary.getMutationCoverage())
-                .build();
-        return view;
+        public double getMutationCoverage() {
+            bot.waitUntil(new ICondition() {
```
echebbi, line 71: Replace `ICondition` with
`org.eclipse.swtbot.swt.finder.waits.DefaultCondition`. You won't have to implement `init` anymore.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/ResultsParser.java
```diff
@@ -56,8 +61,8 @@ public double getMutationCoverage() {
 
     private final String html;
 
-    public ResultsParser(String html) {
-        this.html = html;
+    public ResultsParser(File file) throws IOException {
+        this.html = new String(Files.readAllBytes(Paths.get(file.getAbsolutePath())));
```
echebbi, line 65: `Paths.get(file.getAbsolutePath())` can be replaced with `file.toPath()`.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/AbstractPitclipseSWTBotTest.java
```diff
@@ -121,6 +121,19 @@ public void run() {
...
+            public void run() {
+                IWorkbench workbench = PlatformUI.getWorkbench();
+                try {
+                    workbench.getActiveWorkbenchWindow().getActivePage().showView(viewId);
+                } catch (WorkbenchException e) {
+                    e.printStackTrace();
+                }
+            }
+        });
+    }
```
echebbi, line 135: You can use a lambda expression here:
```java
Display.getDefault().syncExec(() -> {
    ...
});
```

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitSummaryViewTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.pitest.pitclipse.ui.behaviours.pageobjects.PitSummaryView;
+import org.pitest.pitclipse.ui.view.PitView;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.endsWith;
+import static org.hamcrest.Matchers.equalTo;
+import static org.pitest.pitclipse.ui.behaviours.pageobjects.PageObjects.PAGES;
+
+@RunWith(SWTBotJunit4ClassRunner.class)
+public class PitclipsePitSummaryViewTest extends AbstractPitclipseSWTBotTest {
```
echebbi, line 19: A few more scenarios could be tested:

 - making sure the back, forward and home buttons don't break when the view is empty (no mutation has been run)
 - clicking back before clicking on a link
 - clicking forward before clicking on back

echebbi (changes requested):

echebbi:
> Is there any way to know which version of Eclipse and which Browser is used on the server site?

@JKutscha are you talking about the CI server? It should use the versions defined in the target plaform. Those should be the same as when you run Pitclipse from your IDE if your [configured it with the TP](https://github.com/pitest/pitclipse/blob/master/CONTRIBUTING.md#how-do-i-open-the-projects-in-eclipse-ide). And regarding the browser, according to the documentation SWT embeds Safari on macOS.

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
JKutscha: Will fix.

I installed Checkstyle, but it shows more than 400 warnings for files I never touched, and after trying unsuccessfully to solve the problem, I decided to uninstall it.

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
JKutscha: I made it public to access while testing and find it reliable, even if the text is changed.

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
JKutscha: see above.

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
JKutscha: see above.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
JKutscha: Sure, but why is this better than using the bot?

But I will implement it anyway now.

JKutscha: And I also need the bot to change the url with JavaScript in order to be able to use relative urls.

I would prefer a different solution than JavaScript, but I didn't figure one out yet. < Probably a reason why the tests fail sometimes.

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
JKutscha: If we aim for an implementation without a bot, it would be not necessary.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitResultNotifier.java
JKutscha: Thanks.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitSummaryViewTest.java
```diff
@@ -47,16 +47,23 @@ public void navigateWithButtons() throws CoreException {
...
+            assertThat(summaryView.clickBack(), equalTo(lastUrl));
 
             // coverageReportGenerated needs normal timeout
             SWTBotPreferences.TIMEOUT = timeout;
             runPackageTest(FOO_BAR_PACKAGE, TEST_PROJECT);
             coverageReportGenerated(2, 80, 0);
 
             assertThat(summaryView.getCurrentBrowserUrl(), endsWith(INDEX));
-            assertThat(summaryView.setLink("./"+ FOO_BAR_PACKAGE_RESULT+"/"+FOO_CLASS_RESULT), endsWith(FOO_CLASS_RESULT));
+            assertThat(summaryView.setUrl(summaryView.getCurrentBrowserUrl().replace("/index.html",
```
JKutscha, line 59: A bit of a hack right now, but even this does not work for the tests.
If I debug it, the browser shows the correct page, but it fails anyway.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitSummaryViewTest.java
echebbi: @JKutscha I re-open because I don't see how this point is resolved. Please tell me if you think those scenarios are already addressed or explain if you believe that they don't need to be tested.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
```diff
@@ -16,34 +16,87 @@
...
     }
 
-    public void waitForUpdate() {
-        try {
-            lastResults = PitNotifier.INSTANCE.getResults();
-        } catch (InterruptedException e) {
-            throw new StepException(e);
+    public void getViewIfNotSet() {
+        if (summaryView == null) {
+            summaryView = (PitView) Views.getViewById(PitView.VIEW_ID);
```
echebbi, line 37: Could you add a try/catch that logs an error in case the cast to `PitView` fails?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
echebbi: Thank you very much for the change!

> Sure, but why is this better than using the bot?

That brings better cohesion because the methods that modify the _Mutations_ view are now in the class that is responsible of this view. Some advantages I can think of are that:

 - execution is likely (a bit) quicker as we directly trigger an action instead of asking a slow bot to find then click on a button,
 - `PitSummaryView` is no longer dependent on the _Mutations_ view's UI (the bot was assuming that the view was providing its actions as toolbar buttons, which may not be the case in the future),
 - if someone (can be us) wants to reuse `PitView` in the future then they can call, e.g., `back()` or `forward()` right away without having to implement another bot.

JKutscha: Thanks for the explanation. Is this resolved?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
JKutscha: Sure

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitSummaryViewTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Test
+    public void navigateWithButtons() throws CoreException {
+        final long timeout = SWTBotPreferences.TIMEOUT;
+        try {
+            // set timeout to small time, because offline page loads should be quick and
+            // some pages are expected to not change and need timeout
+            SWTBotPreferences.TIMEOUT = 50;
+            // should not change page
+            String lastUrl = summaryView.getCurrentBrowserUrl();
```
JKutscha, line 48: >
>making sure the back, forward and home buttons don't break when the view is empty (no mutation has been run)
@echebbi

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitSummaryViewTest.java
JKutscha: I marked you the point in the test, which I thought resolved this comment.

If this is not sufficient, I am happy to add more or different approaches according to feedback.

LorenzoBettini:
@JKutscha @echebbi I'm afraid I got lost in the conversation :)
I see @echebbi is already reviewing this, so I'm just adding my two cents:

- I personally never use checkstyle and I install SonarLint in Eclipse (which helps me detect issues before SonarCloud reports it)
- please let me know if I can help somehow with the tests... from what I understand testing the browser is not reliable or at least I think we should deal with synchronization issues. Maybe a SWTBot Condition that checks (periodically) the contents of the browser should be enough for having confidence it works (something similar to what I do for checking the contents of the Console waiting for the Console to contain a string saying that PIT has finished)?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
LorenzoBettini: @echebbi sorry if I'm late, but if I understand correctly what you say I think it would be better to actually click on the buttons with the bot, after all, we are testing the UI, aren't we? Or I don't understand what you were suggesting.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
```diff
@@ -36,6 +38,9 @@ public PitSummaryView(SWTWorkbenchBot bot) {
     public void getViewIfNotSet() {
         if (summaryView == null) {
             summaryView = (PitView) Views.getViewById(PitView.VIEW_ID);
+            if (summaryView != null) {
```
LorenzoBettini, line 41: Of course that's a typo: you meant == ;)

JKutscha: Thanks. My head hurt already, from not knowing why it is failing.

LorenzoBettini:
@JKutscha please let me know when this is ready to be reviewed

JKutscha:
> 
> 
> @JKutscha please let me know when this is ready to be reviewed

It is ready.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
JKutscha: I'm open to change this back to use the bot,
Even both ways would be an option for me.

echebbi: @LorenzoBettini you're absolutely right, I completely overlooked the fact that this class aims at testing UI. So, yes, in this context we do want to click on the buttons... 

@JKutscha I'm sorry for the trouble but could you revert back to using the bot? It indeed makes sense to use it here.

JKutscha: Alright, will do

JKutscha: @echebbi Should I leave the methods like `back()` in place or delete them?

JKutscha:
@LorenzoBettini @echebbi I would like to get this merged as quickly as possible, because I want to merge #152 as well and start working on additional features, like #164.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
echebbi: You can delete them. We don't use them right now so they would basically be dead code to maintain.

echebbi (changes requested):
Mostly ask for cosmetic changes. I will also take the time to review the code from the IDE. I mostly want to check the way you changed the synchronization mechanism.

IIRC until now the tests were waiting for the _Mutations_ view to send an event telling it received the results from Pit. Your mechanism waits for mutations results to be available, assumes that the view is updated then immediately moves on. I want to make sure I understand correctly and that there's no synchronization issues (e.g., when the view has a lot of mutations to display).

JKutscha:
> 
> IIRC until now the tests were waiting for the _Mutations_ view to send an event telling it received the results from Pit. Your mechanism waits for mutations results to be available, assumes that the view is updated then immediately moves on. I want to make sure I understand correctly and that there's no synchronization issues (e.g., when the view has a lot of mutations to display).

It waited for the _PitSummary_ (_PitView_) to have the HTML set and parses this HTML source.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
```diff
@@ -16,34 +16,100 @@
...
-    public double getMutationCoverage() {
-        return lastResults.getMutationCoverage();
+    private class BrowserLoadCondition extends DefaultCondition {
+        private final String titleOfPage;
+        private String html;
+        public BrowserLoadCondition(String titleOfPage) {
+            getBrowserIfNotSet();
+            // give the browser time to change.
+            // To avoid false positives, if the page should not change
+            browser.waitForPageLoaded();
```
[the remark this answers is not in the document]
JKutscha: Do you mean to move this in the method of `PitclipsePitSummaryViewTest.java`, if so I disagree?
The browser should wait here, because afterwards the code gets the Url, example from above:
`bot.waitUntil(new BrowserLoadCondition(url.replace(".html", "")));
return getCurrentBrowserUrl();`

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitSummaryViewTest.java
```diff
@@ -0,0 +1,75 @@
...
+            SWTBotPreferences.TIMEOUT = 50;
+            // should not change page
+            assertThat(summaryView.clickHome(BLANK_PAGE), equalTo(BLANK_PAGE));
+            assertThat(summaryView.clickForward(BLANK_PAGE), equalTo(BLANK_PAGE));
+            assertThat(summaryView.clickBack(BLANK_PAGE), equalTo(BLANK_PAGE));
+
+            // coverageReportGenerated needs normal timeout
+            SWTBotPreferences.TIMEOUT = timeout;
+            runPackageTest(FOO_BAR_PACKAGE, TEST_PROJECT);
+            coverageReportGenerated(2, 80, 0);
```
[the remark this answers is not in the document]
JKutscha: This was always the case. Maybe this should be done in a different PR, because this affects almost every ui test.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitSummaryViewTest.java
```diff
@@ -0,0 +1,75 @@
...
+    public void navigateWithButtons() throws CoreException {
+        final long timeout = SWTBotPreferences.TIMEOUT;
+        try {
+            // set timeout to small time, because offline page loads should be quick and
+            // some pages are expected to not change and need timeout
+            SWTBotPreferences.TIMEOUT = 50;
+            // should not change page
+            assertThat(summaryView.clickHome(BLANK_PAGE), equalTo(BLANK_PAGE));
+            assertThat(summaryView.clickForward(BLANK_PAGE), equalTo(BLANK_PAGE));
+            assertThat(summaryView.clickBack(BLANK_PAGE), equalTo(BLANK_PAGE));
```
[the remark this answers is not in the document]
JKutscha: In my testing, `resetView` worked good and it was only added because it was needed for the test running via maven, because the view held the results, as you suspected, even after closing.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitSummaryViewTest.java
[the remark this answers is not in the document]
JKutscha: Also we need to wait for pit in `coverageReportGenerated`. because otherwise the report is not generated.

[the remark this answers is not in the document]
echebbi: Just checked the other tests and, yeah, it will be simpler to address this in another issue. Nevermind.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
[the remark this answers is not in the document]
echebbi: No, I meant in `BrowserLoadCondition::test`.

Current implementation could cause issues, e.g.:
```java
// waits for the page to load here
ICondition condition = new BrowserLoadCondition(url);

// but we change the page here
browser.setUrl(url);

// and thus we may test the page before it's been fully loaded
bot.waitUntil(condition);
```

That's unlikely, I agree, but I see no reason not to fix it, just in case.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitSummaryViewTest.java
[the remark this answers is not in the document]
echebbi: So if I understand you well you're saying that `resetView` actually clears the results from the view, and that it's not the action of closing the view that clears the results? How are the results cleared, in that case?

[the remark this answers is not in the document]
JKutscha: This is correct.
It is not cleared while closing, it is reset while opening. See: `PitView.createPartControl`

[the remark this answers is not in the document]
echebbi: OK, so technically it's not the test that clears the view: the view is _always_ cleared when it opens, even outside our tests. That's not the behavior I'd expect but it looks like it's not a regression so don't bother fixing that in this PR.

So since the view is effectively cleared tests shouldn't interfere with each other. Could you put the asserts into separate test methods?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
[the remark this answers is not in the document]
JKutscha: I agree.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitSummaryViewTest.java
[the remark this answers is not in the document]
JKutscha: Will do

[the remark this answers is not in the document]
LorenzoBettini: > Ok, I missed those new asserts because I was expecting each of them to be into its own test method. It's usually good practice to isolate single test cases because that makes failures easier to identify and tests easier to understand.
> @LorenzoBettini do you have an opinion on these topics?

If I'm still on time answering this, I'm fine with several assertions in the same test method if
- they are related to the same scenario/functionality under test
- they speed up things a bit since the setup is heavy

I would say that this is such a situation: I'd be OK with the 3 asserts in the same test method.

Again, if that has already been changed or you prefer the other way round, forget about what I've just said ;)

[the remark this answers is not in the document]
JKutscha: It is already implemented in a new method.

LorenzoBettini:
@JKutscha @echebbi I've just created a companion PR https://github.com/pitest/pitclipse/pull/169 because I'd like the code to be analyzed by SonarCloud (which cannot currently be done for external PRs, see #161). I basically took the branch of @JKutscha and pushed to this repo (using the same branch name) and created a separate PR. I'd just like to experiment with such a mechanism...

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -16,26 +16,66 @@
...
+    private String homeUrlString = null;
 
     @Override
     public synchronized void createPartControl(Composite parent) {
         try {
             browser = new Browser(parent, SWT.NONE);
+            resetBrowser();
             browser.addProgressListener(new PitUiUpdatePublisher(browser));
+            // create back button
+            final Action backButtonAction = new Action(BACK_BUTTON_TEXT) {
```
LorenzoBettini, line 50: Just cosmetic: since we don't store actions in fields anymore, why not avoiding local variables either and just add to the toolbar using an anonymous class?

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -16,26 +16,66 @@
...
+                }
+            };
+            IActionBars actionBars = getViewSite().getActionBars();
+            IToolBarManager toolBar = actionBars.getToolBarManager();
+            toolBar.add(backButtonAction);
+            toolBar.add(homeButtonAction);
+            toolBar.add(forwardButtonAction);
+            actionBars.updateActionBars();
         } catch (SWTError e) {
             MessageBox messageBox = new MessageBox(parent.getShell(), SWT.ICON_ERROR | SWT.OK);
```
LorenzoBettini, line 80: I know this is not your code but since we're there: `SWTError` is an error and usually you're not supposed to catch errors, which are catastrophic anyway... I'd simply remove the try catch, @echebbi ? This will also avoid uncovered code.

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -44,17 +84,38 @@ public synchronized void createPartControl(Composite parent) {
...
+     */
+    private void resetBrowser() {
+        browser.setUrl(BLANK_PAGE);
+        homeUrlString = null;
+    }
+
     @Override
     public void setFocus() {
-        // Nothing to do
+        if (browser != null && !browser.isDisposed()) {
```
LorenzoBettini, line 99: If understand correctly (we've already talked about that a few days ago), unless something really bad happened, the browser cannot be null, so I'd avoid such a check, which is also uncovered (https://sonarcloud.io/component_measures?id=org.pitest%3Aorg.pitest.pitclipse&metric=new_coverage&pullRequest=169&selected=org.pitest%3Aorg.pitest.pitclipse%3Abundles%2Forg.pitest.pitclipse.ui%2Fsrc%2Forg%2Fpitest%2Fpitclipse%2Fui%2Fview%2FPitView.java&view=list)

while I guess the browser can already be disposed? I see that condition seems to be covered...

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -44,17 +84,38 @@ public synchronized void createPartControl(Composite parent) {
...
-            browser.setUrl(result.toURI().toString());
+            homeUrlString = result.toURI().toString();
+            browser.setUrl(homeUrlString);
+        }
+    }
+
+    @Override
+    public void dispose() {
+        if (browser != null && !browser.isDisposed()) {
+            browser.dispose();
```
LorenzoBettini, line 117: similar to the previous comment, but this time `browser.dispose` does not seem to be covered by tests... I wonder why...

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
```diff
@@ -16,34 +16,97 @@
...
+import org.eclipse.swtbot.swt.finder.waits.DefaultCondition;
+import org.eclipse.swtbot.swt.finder.widgets.SWTBotBrowser;
+import org.pitest.pitclipse.ui.view.PitView;
 
 public class PitSummaryView {
+    private SWTWorkbenchBot bot;
+    private SWTBotBrowser browser;
 
-    private PitResultsView lastResults = null;
+    public PitSummaryView(SWTWorkbenchBot bot) {
```
LorenzoBettini, line 28: Should I also review this test or shall we wait for the PR https://github.com/pitest/pitclipse/pull/167 to be merged?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/Views.java
```diff
@@ -115,4 +119,18 @@ public SWTBotView showConsole() {
...
     }
 
+
+    /**
+     * Gets the view specified by the given id, if the view is open, otherwise returns null.
+     * @param viewId of view will get returned, if open
+     * @return view or null, if not found
+     */
+    public static IViewPart getViewById(String viewId) {
+        AtomicReference<IViewPart> view = new AtomicReference<>();
```
LorenzoBettini, line 129: Why not using SWTBot here? If I understand correctly this is not about opening a view, but finding a view right?
I would expect an iteration on current views provided by SWTBot till you find the needed one. Something similar is done somewhere else in our UI tests.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/steps/PitclipseSteps.java
```diff
@@ -60,24 +59,24 @@
...
         runPit(new SelectTestClass(testClassName, packageName, projectName));
     }
 
     @Then("a coverage report is generated with {int} class/classes tested with overall coverage of {int}% and mutation coverage of {int}%")
     public void coverageReportGenerated(int classes, double totalCoverage, double mutationCoverage) {
-        PitSummaryView pitView = PAGES.getPitSummaryView();
-        pitView.waitForUpdate();
+        PAGES.views().waitForTestsAreRunOnConsole();
```
LorenzoBettini, line 74: I guess I should wait before reviewing this right? This should be related to https://github.com/pitest/pitclipse/pull/167

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/AbstractPitclipseSWTBotTest.java
```diff
@@ -134,6 +136,31 @@ public void run() {
...
+            try {
+                workbench.getActiveWorkbenchWindow().getActivePage().showView(viewId);
+            } catch (WorkbenchException e) {
+                e.printStackTrace();
+            }
+        });
+    }
+
+    public static void closeViewById(String viewId) throws InterruptedException {
+        Display.getDefault().syncExec(() -> {
```
LorenzoBettini, line 151: similar to what I said above: while opening a view programmatically is good, closing it could be easily done with SWTBot by iterating over the current views...

LorenzoBettini (changes requested):
As you see in the comments, I guess I should not review the tests since when https://github.com/pitest/pitclipse/pull/167 is merged this PR should be updated as well, is that right?

Besides that, I have a few concerns about uncovered lines (see the comments), see also https://sonarcloud.io/component_measures?id=org.pitest%3Aorg.pitest.pitclipse&metric=new_coverage&pullRequest=169&selected=org.pitest%3Aorg.pitest.pitclipse%3Abundles%2Forg.pitest.pitclipse.ui%2Fsrc%2Forg%2Fpitest%2Fpitclipse%2Fui%2Fview%2FPitView.java&view=list

Thank you!

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
echebbi: Not sure about that. I agree that errors shouldn't usually being caught but in this case it may make sense. `Browser`'s constructor is roughtly implemented as follows:

```java
this.webBrowser = BrowserFactory.createWebBrowser();
if (this.webBrowser == null) {
    // factory was unable to load WebKit
    throw new SWTError();
}
```

I guess the constructor throws an error to report that the issue comes from the execution environment (a library is missing) but I believe that it makes sense to catch the error so that we can report it nicely to the user. I just tried to broadcast the error and it's basically swallowed: some log is added to the _Error Log_ view and that's it. Not really helpful for a lambda user.

Actually, I believe that we should do better reporting:

- the MessageBox should say that it cannot open a web browser to display Pitest's HTML report (otherwise user may wonder why the view even tries to open a browser),
- the MessageBox should report the error's message to help troubleshooting,
- the PitView could display an error message instead of just remaining blank.

Does that sound reasonable? If you agree with that I've written then we should address this in another PR.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/steps/PitclipseSteps.java
JKutscha: Yes, please.

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
JKutscha: Will do.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
JKutscha: Please wait for #167.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/Views.java
JKutscha: Will be gone after #167

JKutscha:
@LorenzoBettini @echebbi please review this PR, if you find some time. From my part, this PR is ready.

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -18,28 +18,75 @@
...
+                },
+                parent,
+                "Failed to create the Pit Summary view",
+                "Browser cannot be initialized.");
+    }
+
+    /**
+     * Set URL to blank and reset homeUrl, to have a clean browser after
+     * creation<br>
+     * <b>Needed</b> for testing purposes
```
LorenzoBettini, line 85: This is not needed anymore by tests, isn't it?
Indeed it's private

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -18,28 +18,75 @@
...
-        PitclipseUiUtils.executeSafely
-            (() -> {
-                browser = new Browser(parent, SWT.NONE);
-                browser.addProgressListener(new PitUiUpdatePublisher(browser));
-            },
-            parent,
-            "Failed to create the Pit Summary view",
-            "Browser cannot be initialized.");
+        PitclipseUiUtils.executeSafely(
+                () -> {
```
LorenzoBettini, line 45: just cosmetic: we could take an indent back (4 spaces should be enough for indentation, instead of 8)

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
```diff
@@ -0,0 +1,112 @@
...
+        getBrowserIfNotSet();
+        browser.setUrl(url);
+        bot.waitUntil(new BrowserLoadCondition(url.replace(".html", "")));
+        return getCurrentBrowserUrl();
+    }
+
+    private class BrowserLoadCondition extends DefaultCondition {
+        private final String titleOfPage;
+        private String html;
+        public BrowserLoadCondition(String titleOfPage) {
```
LorenzoBettini, line 83: just cosmetic: please add an empty line before the constructor

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/PitSummaryView.java
```diff
@@ -0,0 +1,112 @@
...
+            getBrowserIfNotSet();
+            final int lastSegment = titleOfPage.lastIndexOf('/') + 1;
+            if (lastSegment > 0) {
+                this.titleOfPage = titleOfPage.substring(lastSegment);
+            } else {
+                this.titleOfPage = titleOfPage;
+            }
+        }
+        @Override
```
LorenzoBettini, line 92: just cosmetic: please add an empty line before the method

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/AbstractPitclipseSWTBotTest.java
```diff
@@ -132,6 +135,30 @@ public void run() {
...
+                e.printStackTrace();
+            }
+        });
+    }
+
+    public static void closeViewById(String viewId) {
+        final long previousTimeout = SWTBotPreferences.TIMEOUT;
+        try {
+            // set timeout short, because if the view is not present we don't wait long
+            SWTBotPreferences.TIMEOUT = 100;
```
LorenzoBettini, line 153: Please, iterate over the views using SWTBot and close the view if found.
Much better than this strategy, which is much less deterministic.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitSummaryViewTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @BeforeClass
+    public static void setupJavaProject() throws CoreException {
+        importTestProject(TEST_PROJECT);
+    }
+
+    @Before
+    public void resetView() throws InterruptedException {
+        // If this class is run alone, the view can't be closed and a RuntimeException
+        // is thrown as expected
```
LorenzoBettini, line 37: I'm not sure I understand this comment: do you mean that this test cannot be run alone?

LorenzoBettini (changes requested):
I'm basically fine, but I left a few (cosmetic) comments, some questions, and one crucial change to be done.

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
JKutscha: This is still needed.
It is called while creating and if an `update()` has no file.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitSummaryViewTest.java
JKutscha: Is gone now, was connected with `closeViewById()`

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
LorenzoBettini: Directly from the tests? By reflection?

JKutscha: No, in the methods `update(...)` and `createPartControl(...)`.

LorenzoBettini (approved):
I'm fine with that... I only have a doubt on that Javadoc on the private method, but probably it's just my misunderstanding (when I read "Needed for tests" I mean a method, typically package-private, that is meant to be called directly by tests...)

@echebbi , if you're fine with the PR please merge it

LorenzoBettini:
@echebbi can I merge this or do you want to further review it?

echebbi (approved):

echebbi:
> @echebbi can I merge this or do you want to further review it?

@LorenzoBettini I don't have the time to review it right now but @JKutscha addressed my previous comments so if it looks good to you please proceed

PR #153, opened 2021-07-22
Macos experiments

LorenzoBettini:
This fixes the problems we started to have on macOS GitHub Actions virtual environments.
It updates Eclipse TP to 2020-06 the last one NOT requiring Java 11. (I also fixed a few deprecated warnings)

@echebbi I really think that we should start building against the latest released Eclipse platform. That will require Java 11 but we already use Java 11 for SonarCloud. Moreover, it will NOT require Pitclipse users to use Java 11 anyway. This will also allow us to use Tycho more up-to-date versions, benefit from bug fixes and deal with problems we started to experience in the CI. I will create a separate issue for that.

@JKutscha when this gets merge please update your current working branches.

### DIFF
--- a/.github/workflows/windows-mac.yml
+++ b/.github/workflows/windows-mac.yml
@@ -21,10 +21,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 1.8
+    - name: Set up JDK 11
       uses: actions/setup-java@v1
       with:
-        java-version: 1.8
+        java-version: 11
     - name: Cache Maven packages
       uses: actions/cache@v2
       with:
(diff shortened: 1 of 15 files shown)


LorenzoBettini:
@echebbi I'm merging this since otherwise builds on macOS won't work

echebbi:
@LorenzoBettini I just want to make sure I understand the consequences of building against Eclipse 2021-06. As long as we don't change the version ranges of required bundles in the _MANIFEST.MF_ files then Pitclipse will remain compatible with the same environment as before, won't it? In other words, building against latest Eclipse release won't prevent Pitclipse from running in older Eclipse environments and hence will change nothing for the end users, right?

LorenzoBettini:
@echebbi exactly as you said!

Of course, we won't be sure that Pitclipse will work also on older versions of Eclipse...

Moreover, we have to be careful not to use a new API that's not available in previous versions of Eclipse.

What we can do is to have a CI build that builds against an old version of Eclipse (e.g., Photon), maybe only on Linux. That build will make sure that we don't use a new API (otherwise it won't compile). We could also run the core tests. Of course, we cannot easily run UI tests since we might experience problems like the one on macOS. Thus UI tests will not be run in such a build. In any case, at some point, we'll have to drop older versions of Eclipse or it will become a problem to maintain.

What do you think?

LorenzoBettini:
I created https://github.com/pitest/pitclipse/issues/154 for that

PR #156, opened 2021-07-23
Task 154 build using the latest eclipse

### LINKED ISSUE #154: Build using the latest Eclipse platform
As shown here https://github.com/pitest/pitclipse/pull/153 using an old Eclipse TP might get us into trouble in the CI, due to incompatibilities of older versions of Eclipse (mainly SWT) with newer versions of OSes.

We should build against the latest released Eclipse platform, so that we can also benefit from latest bug fixes and improvements.

To make sure Pitclipse is still compatible with older Eclipse versions we should also have a CI workflow that uses a minimal required version, e.g., Photon. In such a workflow we only run core tests and not UI tests that might be unreliable with such old versions of Eclipse. We could build only in Linux in that case.

LorenzoBettini:
Closes #154

### DIFF
--- a/.github/workflows/maven-old-tp.yml
+++ b/.github/workflows/maven-old-tp.yml
@@ -1,0 +1,54 @@
+# It builds using an old Eclipse TP
+
+name: Java CI with Maven Old TP
+
+on:
+  push:
+    paths-ignore:
+     - 'README.md'
+  pull_request:
+    paths-ignore:
+     - 'README.md'
+
+jobs:
+  build:
+    env:
+      SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        # Shallow clones should be disabled for a better relevancy of SonarQube analysis
+        fetch-depth: 0
+    - name: Set up JDK 11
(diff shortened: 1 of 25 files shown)


LorenzoBettini:
@echebbi as you can see, we also have a build that uses the old TP Photon, while the other builds all use the new TP.
Currently UI tests with Photon are still run (under Linux), if they start to fail we'll disable UI tests in the build that uses the old TP.

.github/workflows/maven-old-tp.yml
```diff
@@ -0,0 +1,54 @@
...
+
+name: Java CI with Maven Old TP
+
+on:
+  push:
+    paths-ignore:
+     - 'README.md'
+  pull_request:
+    paths-ignore:
+     - 'README.md'
```
echebbi, line 11: Maybe ignore all markdowns (`**.MD`) to prevent builds when modifying the contribution guide and the changelog?

echebbi:
> @echebbi as you can see, we also have a build that uses the old TP Photon, while the other builds all use the new TP.

@LorenzoBettini Perfect, thank you!

.github/workflows/maven-old-tp.yml
LorenzoBettini: right! I'll do that!

PR #160, opened 2021-07-27
Task 136 get rid of guava

### LINKED ISSUE #136: Get rid of Guava?
## Motivation

Do we really need Guava? With Java 8 we could do without.
While working on #135 I have already gotten rid of Guava Optional (in favor of Java 8 Optional, as reported by Sonar); similarly for Guava Object that can be replaced by Java Objects.

We could also do without Immutable Collections, couldn't we? Do we really need Immutable Collections properties?

My main concerns about our use of Guava is that in future versions of Eclipse a newer version of Guava might be present that is not more compatible with the one we use

LorenzoBettini:
Closes #136


LorenzoBettini:
@echebbi @JKutscha I'd be ready to merge this.

PR #166, opened 2021-07-29
165: settings for ignoring the warning

### LINKED ISSUE #165: Java dependencies don't match
grafik

As you can see above, we define Java 11 as a dependency, but in the other plug ins we use Java 8.
I would like to up everything to Java 11.

LorenzoBettini:
Closes #165 

Task-Url: http://github.com/pitest/pitclipse/issues/165


LorenzoBettini:
@echebbi this is a quick fix for ignoring the warnings

PR #152, opened 2021-07-20
Mutator configuration

JKutscha:
Issue #114
grafik

PR #167, opened 2021-07-29
Restructered tests to not rely on the Summary View

JKutscha:
Removed two `coverageReportGenerated()` where no report is generated in the newest PIT version


JKutscha:
@LorenzoBettini @echebbi this is the requested PR from #152, please review this if you have time.

tests/org.pitest.pitclipse.ui.tests/fragment.xml
```diff
@@ -1,16 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.4"?>
 <fragment>
-   <extension
```
LorenzoBettini, line 4: Just for curiosity: here only the order of the extensions has changed... is that important?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitResultNotifier.java
```diff
@@ -16,36 +16,75 @@
...
 
-    private void notifiyTestsOfHtmlResults(PitUiUpdate updateEvent) {
-        PitResultsView view = buildResultsView(updateEvent);
-        tryNotifyResults(view);
-    }
+        private Summary summary;
+        private SWTWorkbenchBot bot;
+
+        public int getClasses() {
+            bot.waitUntil(new SummaryWaitCondition(), SWTBotPreferences.TIMEOUT);
```
LorenzoBettini, line 39: I'm not sure I like this new strategy: before we were waiting for the summary to be ready and then we were accessing the results. Now, we don't explicitly wait for the summary to be ready but we implicitly wait for it when retrieving classes AND coverage AND mutation coverage (of course, only the first wait will be a real wait, but still...)... since we must be aware of resetting, why shouldn't be aware also of the wait before accessing the results? Or am I missing something?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipseUiRunnerTest.java
```diff
@@ -34,7 +34,6 @@ public void emptyClassAndEmptyTest() throws CoreException {
         runTest(FOO_TEST_CLASS, FOO_BAR_PACKAGE, TEST_PROJECT);
         consoleContains(0, 0, 100, 0, 0);
         mutationsAre(Collections.emptyList());
-        coverageReportGenerated(0, 100, 100);
```
LorenzoBettini, line 37: From what I understand you removed these calls because with the new PIT in this case no HTML is generated right? But what happens if you leave these calls (that is, the developer inadvertently calls that or a test that was succeeding now fails because of a breaking change in the code)? Exception? Immediate failure? Timeout?

LorenzoBettini (changes requested):
Thank you for isolating the changes :)

So if I understand correctly, this PR cleans up things a bit, by removing an intermediate layer, but without changing the semantics of the test right? (Besides the last comment).

Basically, this is not strictly required to make tests succeed or it's required for the other PRs of yours? Again, just to understand things better, NOT because I'm against it ;)

In general, it looks fine to me, but as you see I have a few doubts about some parts.

tests/org.pitest.pitclipse.ui.tests/fragment.xml
JKutscha: No, the actual point changed from:
org.pitest.pitclipse.**ui**.results
to:
org.pitest.pitclipse.**core**.results

This is one of the most important changes, because it changes the time and source where this class gets notified.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitResultNotifier.java
JKutscha: This class is not responsible for waiting for pit to finish. These methods are only called at one place and prior to that we are waiting for pit to finish. These waits are only in place to wait for a short while to make up for the desynchronization between the extension point and our waiting mechanism.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipseUiRunnerTest.java
JKutscha: They run into a timeout, if they were left in place.

>
>[...] or a test that was succeeding now fails because of a breaking change in the code)? Exception? Immediate failure? Timeout?
I don't understand what you mean with that.

JKutscha:
> 
> So if I understand correctly, this PR cleans up things a bit, by removing an intermediate layer, but without changing the semantics of the test right? (Besides the last comment).

It fixes an issue which causes test to fail, if they are running longer than the `SWTBotPreference.TIMEOUT`, which by default is set to 5 seconds.

Also, why I reworked it in the first place, no tests break, if the pit summary view is changed in any way. Prior the html of the pit summary view was used to get the html text and parse it. If we decide to change the view, looking at synchronization I proposed or deleting the view (Editor highlighting instead of this view), all tests will remain fine. Would you delete this view currently it would break every other test which uses `coverageReportGenerated`. 

> 
> Basically, this is not strictly required to make tests succeed or it's required for the other PRs of yours? Again, just to understand things better, NOT because I'm against it ;)

As stated above, I think it is required to have better tests and be less dependent on things which are not related with the tests at all. Also, yes it is needed for my PRs.

JKutscha:
I now made a bigger overhaul with waiting for pit in general. `RunPit` now is `runPitAndWaitForIt` which imo makes a lot more sense, because we don't parse the console output for the result, which could change in the future with pit.

@echebbi maybe this is what you thought of in this [comment](https://github.com/pitest/pitclipse/pull/150#discussion_r678873927).
@LorenzoBettini [This](https://github.com/pitest/pitclipse/pull/167#discussion_r679101925) is now solved with a separate call to `noCoverageReportGenerated`. Also [this](https://github.com/pitest/pitclipse/pull/167#discussion_r679099194) is solved.

JKutscha:
The failures I'm seeing are related with `consoleContains`, which I personally dislike, because as mentioned before the output of the console can change in the future. I will try to get this check working differently.

LorenzoBettini:
> The failures I'm seeing are related with `consoleContains`, which I personally dislike, because as mentioned before the output of the console can change in the future. I will try to get this check working differently.

I'm not a big fan of consoleContains either though I personally introduced that O:)
I did that when I wrote from scratch many UI tests with SWTBot (without using Cucumber, though we're still use some old code of the Cucumber steps... we'll have to get rid of Cucumber dependencies at some point and remove lots of unused testing code).

I introduced that as a first mechanism for checking whether PIT ran correctly and as expected because I did not want to use the browser in the Pit View to do that.

If we find a better way of checking the results of PIT without using any of our UI views in some specific tests then I'd be very happy. Maybe as you started to do: wait for notification from PIT and parse the results of HTML? That way, we'd be able to test that our run configurations work correctly independently from the PIT View tree and from the PIT View browser. Of course, we still need to test that the tree view works correctly (as we already do by double-clicking on some nodes and verifying that the corresponding editor is opened) and that the browser shows the expected contents. But, as I said, I agree that for verifying PIT runs we shouldn't need to use the Console (by the way, when migrating to PIT 1.6.7 those tests broke and had to be adapted to the new output). @echebbi what do you think?

JKutscha:
> 
> [...] If we find a better way of checking the results of PIT without using any of our UI views in some specific tests then I'd be very happy. Maybe as you started to do: wait for notification from PIT and parse the results of HTML? [...]

@LorenzoBettini If you can live without the test on `testsRun` and `testsPerMutations` of the method
`consoleContains(int generatedMutants, int killedMutants,
            int killedPercentage,
            int testsRun,
            int testsPerMutations)` we could easily parse them from the HTML file.

LorenzoBettini:
> > [...] If we find a better way of checking the results of PIT without using any of our UI views in some specific tests then I'd be very happy. Maybe as you started to do: wait for notification from PIT and parse the results of HTML? [...]
> 
> @LorenzoBettini If you can live without the test on `testsRun` and `testsPerMutations` of the method
> `consoleContains(int generatedMutants, int killedMutants, int killedPercentage, int testsRun, int testsPerMutations)` we could easily parse them from the HTML file.

Yes, I think we can definitely live without them... it was just a further check I was doing at that time, but I don't think they're required to verify that we can actually run PIT

echebbi:
> @echebbi maybe this is what you thought of in this comment.

@JKutscha Yes, mostly. I'm a bit worried by the approach though. If I understand your implementation correctly, `runPitAndWaitForIt` makes UI tests wait for Pitest to end running before moving on. Tests are "unblocked" when `PitResultNotifier` receives the results sent by `PitRunnerMain`. However, since views receive the results at the same time and then take some time to update, we might run into cases where the tests move on _before_ the views are updated, leading to unexpected failures. Am I mistaken?

> But, as I said, I agree that for verifying PIT runs we shouldn't need to use the Console (by the way, when migrating to PIT 1.6.7 those tests broke and had to be adapted to the new output). @echebbi what do you think?

@LorenzoBettini I agree that if we just want to make sure Pitest run and sent the expected results then we shouldn't have to read the console. Additional tests checking that Pitest correctly prints on the console could still be useful though, I don't really have a strong opinion on that.

> >[...] If we find a better way of checking the results of PIT without using any of our UI views in some specific tests then I'd be very happy. Maybe as you started to do: wait for notification from PIT and parse the results of HTML? [...]
>
> @LorenzoBettini If you can live without the test on testsRun and testsPerMutations of the method
consoleContains(int generatedMutants, int killedMutants, int killedPercentage, int testsRun, int testsPerMutations) we could easily parse them from the HTML file.

Just so you know we don't have to manually parse the HTML file.

Firstly, a HTML parser is already implemented: [ResultsParser](https://github.com/pitest/pitclipse/blob/6f6994f47a8de5f28523500a7c03b40203b5a1b5/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/ResultsParser.java#L25) which use is demonstrated in [PitResultNotifier](https://github.com/pitest/pitclipse/blob/6f6994f47a8de5f28523500a7c03b40203b5a1b5/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitResultNotifier.java#L35-L36).

Secondly, [`PitRunnerMain` sends a `PitResults` object when Pitest is done](https://github.com/pitest/pitclipse/blob/6f6994f47a8de5f28523500a7c03b40203b5a1b5/bundles/org.pitest.pitclipse.runner/src/org/pitest/pitclipse/runner/PitRunnerMain.java#L39-L41); this object contains all the mutations that have been detected by Pitest and we can use it to check PIT results. [`PitResultsNotifier`](https://github.com/pitest/pitclipse/blob/6f6994f47a8de5f28523500a7c03b40203b5a1b5/bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/notify/PitResultsNotifier.java#L35) already catches this object (it uses it to feed the summary view) so we should be able to retrieve it from the tests with just minor changes.

JKutscha:
> 
> 
> > @echebbi maybe this is what you thought of in this comment.
> 
> @JKutscha Yes, mostly. I'm a bit worried by the approach though. If I understand your implementation correctly, `runPitAndWaitForIt` makes UI tests wait for Pitest to end running before moving on. Tests are "unblocked" when `PitResultNotifier` receives the results sent by `PitRunnerMain`. However, since views receive the results at the same time and then take some time to update, we might run into cases where the tests move on _before_ the views are updated, leading to unexpected failures. Am I mistaken?

I think this is not unexpected and should be managed by the tests which need to wait for specific views, because otherwise we would make an approach which waits longer than it needs to. Also, we didn't do it different before. Now it is always the same and we don't wait indirectly for the tests. This approach is IMHO more straightforward and is more consistent.

> 
> Just so you know we don't have to manually parse the HTML file.

I looked into the parser and I'm not sure if this is the best way, if we could use the result object instead.

> [...] Secondly, [`PitRunnerMain` sends a `PitResults` object when Pitest is done](https://github.com/pitest/pitclipse/blob/6f6994f47a8de5f28523500a7c03b40203b5a1b5/bundles/org.pitest.pitclipse.runner/src/org/pitest/pitclipse/runner/PitRunnerMain.java#L39-L41); this object contains all the mutations that have been detected by Pitest and we can use it to check PIT results. [`PitResultsNotifier`](https://github.com/pitest/pitclipse/blob/6f6994f47a8de5f28523500a7c03b40203b5a1b5/bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/notify/PitResultsNotifier.java#L35) already catches this object (it uses it to feed the summary view) so we should be able to retrieve it from the tests with just minor changes.

I will try to retrieve all the necessary information from the results directly, without parsing.

JKutscha:
> 
> I will try to retrieve all the necessary information from the results directly, without parsing.

Sadly, line coverage and mutation coverage can only be parsed. I deleted `HTMLTable`, because I think my approach using regex is more compact and still easy to understand.


Regarding my failed tests from the CI, I tested them all locally and they run well, but yesterday I couldn't run maven locally because of some dependency which could not be accessed. It seems this still is the case today, at least in some cases.

From my part, I think this PR is finished.

@LorenzoBettini @echebbi I kindly ask you guys to review this PR, thanks :)

echebbi:
> @LorenzoBettini @echebbi I kindly ask you guys to review this PR, thanks :)

Thank you for your work @JKutscha :) I'll take the time to review your code today! I just have a few comments/questions beforehand.

> Regarding my failed tests from the CI, I tested them all locally and they run well, but yesterday I couldn't run maven locally because of some dependency which could not be accessed. It seems this still is the case today, at least in some cases.

Indeed, I don't know what's going on but some old Eclipse p2 update sites are unavailable since yesterday. I guess that it should be fixed soon.

> > > @echebbi maybe this is what you thought of in this comment.
> > 
> > 
> > @JKutscha Yes, mostly. I'm a bit worried by the approach though. If I understand your implementation correctly, `runPitAndWaitForIt` makes UI tests wait for Pitest to end running before moving on. Tests are "unblocked" when `PitResultNotifier` receives the results sent by `PitRunnerMain`. However, since views receive the results at the same time and then take some time to update, we might run into cases where the tests move on _before_ the views are updated, leading to unexpected failures. Am I mistaken?
> 
> I think this is not unexpected and should be managed by the tests which need to wait for specific views, because otherwise we would make an approach which waits longer than it needs to.

Yes, tests should wait for views to be updated but in your implementation they don't, do they? 

> Also, we didn't do it different before.

We were [waiting for the UI in PitclipseSteps::coverageReportGenerated](https://github.com/pitest/pitclipse/blob/6f6994f47a8de5f28523500a7c03b40203b5a1b5/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/steps/PitclipseSteps.java#L76). There's a synchronization mechanism thanks to the _org.pitest.pitclipse.ui.results_ extension point: basically, the view sends an event when it's updated and `PitSummaryView` waits for it.

I'm not saying that `coverageReportGenerated` is the right place to wait but I believe that we should make sure that tests are paused until the corresponding view is fully updated.

> Sadly, line coverage and mutation coverage can only be parsed.

I'm fine with them being parsed for now. Just for the record, in the future we may want to check how Pitest generates its HTML reports: we may be able to retrieve the coverages without having to depend on the HTML report. The corresponding class seems to be [MutationHtmlReportListener](https://github.com/hcoles/pitest/blob/e2c1a2324253b9c8d5bb57c3568079d84b3780eb/pitest-html-report/src/main/java/org/pitest/mutationtest/report/html/MutationHtmlReportListener.java).

> I deleted HTMLTable, because I think my approach using regex is more compact and still easy to understand.

Once again you're starting to refactor a lot of things in a single PR :) Is there any functional reason to refactor this class? It looks like you need two new properties (`generatedMutants` and `killedMutants`) but you get them from `PitResults` so I don't think a rewrite was needed. If it's just a cosmetic refactor I'd prefer to have it addressed in another PR. 

And by the way, if you only wanted to change the implementation, why did you removed the whole `HTMLTable` class? Now that `getRows`, `getHeaders`, etc. methods are in `ResultsParser` the code looks less clear to me.

JKutscha:
> 
> Yes, tests should wait for views to be updated but in your implementation they don't, do they?

They kind of do, because we are using the SWTBot. The bot itself waits for things to appear. In our case is done while getting the tree, which holds the mutators int the `MutationView`

> 
> We were [waiting for the UI in PitclipseSteps::coverageReportGenerated](https://github.com/pitest/pitclipse/blob/6f6994f47a8de5f28523500a7c03b40203b5a1b5/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/steps/PitclipseSteps.java#L76). There's a synchronization mechanism thanks to the _org.pitest.pitclipse.ui.results_ extension point: basically, the view sends an event when it's updated and `PitSummaryView` waits for it.

You are correct, but I didn't count that because we were waiting for the `PitSummaryView` which currently is not tested in any way. The view we test is the `MutationView` and we wait there, mentioned above.

>
> Once again you're starting to refactor a lot of things in a single PR :) Is there any functional reason to refactor this class? It looks like you need two new properties (`generatedMutants` and `killedMutants`) but you get them from `PitResults` so I don't think a rewrite was needed. If it's just a cosmetic refactor I'd prefer to have it addressed in another PR.

I added these new properties and wanted them to be in the summary, so the refactor was necessary. Maybe I don't understand what you mean, would you try to explain it to me differently.

> 
> And by the way, if you only wanted to change the implementation, why did you removed the whole `HTMLTable` class? Now that `getRows`, `getHeaders`, etc. methods are in `ResultsParser` the code looks less clear to me.

Maybe I should rename them, but I also could put them back to the `HTMLTable` class. I thought having one parser class would be more clear.

echebbi:
> They kind of do, because we are using the SWTBot. The bot itself waits for things to appear. In our case is done while getting the tree, which holds the mutators int the MutationView

Ok, makes sense.

> I added these new properties and wanted them to be in the summary, so the refactor was necessary. Maybe I don't understand what you mean, would you try to explain it to me differently.

I was talking about refactoring the way the HTML report is parsed. You replaced the existing implementation with a regex-based one; was there any need to?

JKutscha:
> 
> I was talking about refactoring the way the HTML report is parsed. You replaced the existing implementation with a regex-based one; was there any need to?

I refactored the `HTMLTable` because the summary was created there and now it has the two additional fields, which aren't parsed from the HTML. Also I wanted to get all in the parser class.

The parsing inside the `HTMLTable` was working, but I figured if I'm already reworking parts of it, make it less code.

echebbi:
> I refactored the `HTMLTable` because the summary was created there and now it has the two additional fields, which aren't parsed from the HTML.

I'm sorry, that doesn't make sense to me. Summary is created in `ResultsParser` and since the two additional fields aren't parsed from HTML then there's no functional need to refactor `HTMLTable`.

> The parsing inside the HTMLTable was working, but I figured if I'm already reworking parts of it, make it less code.

Maybe, but, since it has nothing to do with decoupling the tests from the Summary view, making the change here just adds noise to your contribution. Refactoring `HTMLTable` brings several questions/comments;

 - does it bring any value?
 - do we really want to get rid of the `HTMLTable` abstraction?
 - regex patterns should be compiled statically instead of within a method
 - the `getTable`, `getRows` and `getHeaders` method may be refactorable as a single method
 - we may want additional tests to prevent regressions
 - etc.

and I believe that this PR already brings a lot of discussion. So if you do think that refactoring this class is worth it then please submit a new PR for it (yes, yet another one). The shorter a PR, the quicker it can be reviewed and merged :)

JKutscha:
>
> [...] and I believe that this PR already brings a lot of discussion. So if you do think that refactoring this class is worth it then please submit a new PR for it (yes, yet another one). The smaller a PR, the quicker it can be reviewed and merged :)

I will revert it back to the original `HTMLTable`.

JKutscha:
> 
> I'm sorry, that doesn't make sense to me. Summary is created in `ResultsParser` and since the two additional fields aren't parsed from HTML then there's no functional need to refactor `HTMLTable`.

@echebbi I reverted it back now and I hope the results good.

JKutscha:
@echebbi @LorenzoBettini IMO this PR is ready for review. I would like to get this merged, that I can work on #150 and #152.

The reason I want to move faster, is that I write my Bachelor Thesis about this work and I need to get my changes merged, that I can introduce Pitclipse in a selected group of people. My deadline isn't so long in the future, so please find some time to review, thanks :)

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/Console.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public Console(SWTWorkbenchBot bot) {
+        this.bot = bot;
+    }
+
+    /**
+     * Shows the view console and returns its text.
+     * @return the console text
+     */
+    public String getConsoleText() {
```
echebbi, line 36: Can be renamed "`getText`" (the "Console" word is redundant because we're already in the `Console` class)

echebbi (changes requested):
Just a minor change on my side

echebbi (approved):

JKutscha:
@LorenzoBettini can we merge?

LorenzoBettini:
I'm going to review it right now, in the meantime I created the companion PR https://github.com/pitest/pitclipse/pull/171 for SonarCloud analysis

LorenzoBettini (approved):
I'm basically fine with that: I haven't reviewed that deeply, but I followed the conversation with @echebbi and I agree with all the comments and addressed comments.

I only left two "doubts" concerning the completely removed code for testing the console.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/Views.java
```diff
@@ -1,118 +0,0 @@
-/*******************************************************************************
```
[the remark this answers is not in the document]
JKutscha: This class now exists as `Console.java`, which imo is more fitting and we can get the text from the console there. This would allow us to check the console for some content in the future.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/AbstractPitclipseSWTBotTest.java
```diff
@@ -309,53 +307,6 @@ protected static void runProjectTest(final String projectName) throws CoreExcept
         new PitclipseSteps().runProjectTest(projectName);
     }
 
-    protected static void consoleContains(int generatedMutants, int killedMutants,
```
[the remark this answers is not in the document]
JKutscha: See above and:

This in particular is now integrated to the most part with `coverageReportGenerated()`.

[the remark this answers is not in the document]
echebbi: I think we'll only need this method if we decide to test what's being printed on the console (which could be useful by the way). This method is pretty straightforward so if it's not used at the moment I'm fine with getting rid of it for now.

LorenzoBettini:
OK, thanks for your answers @JKutscha and @echebbi as soon as the companion PR completes, I'm going to merge this one!

JKutscha:
> 
> 
> OK, thanks for your answers @JKutscha and @echebbi as soon as the companion PR completes, I'm going to merge this one!

Thanks

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/Views.java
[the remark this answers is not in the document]
echebbi: I indeed missed that. @JKutscha it looks like some methods have been lost though (i.e., `clear`, `close` and `show`). Even if they're not used anymore they look like useful methods that we'll likely use in the future, and since they're already implemented we may want to move them to the `Console` class...

I don't have a strong opinion on that but I I'd prefer the methods to be kept because, as I said, they are very likely to be used in the future.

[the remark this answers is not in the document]
JKutscha: Will do.

LorenzoBettini:
Thank you @JKutscha for your contribution, which is now merged!
You may want to delete this branch (using the web interface of this PR)

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/Views.java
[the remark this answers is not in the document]
LorenzoBettini: Opps... sorry I have already merged... I'll wait for the next PR

PR #170, opened 2021-08-02
UI utils

LorenzoBettini:
This is a proposal to address the conversation of https://github.com/pitest/pitclipse/pull/150, in particular, https://github.com/pitest/pitclipse/pull/150#discussion_r680315622

@echebbi @JKutscha what do you think?


LorenzoBettini:
By the way, we get failures with the old TP, because Eclipse download sites have a few problems since yesterday

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -33,20 +32,18 @@
...
-            browser = new Browser(parent, SWT.NONE);
-            browser.addProgressListener(new PitUiUpdatePublisher(browser));
-        } catch (SWTError e) {
-            MessageBox messageBox = new MessageBox(parent.getShell(), SWT.ICON_ERROR | SWT.OK);
-            messageBox.setMessage("Browser cannot be initialized.");
-            messageBox.setText("Exit");
-            messageBox.open();
-        }
+        PitclipseUiUtils.safeExecute
+            (() -> {
```
JKutscha, line 36: Just cosmetic, why is this in a new line?

JKutscha (changes requested):

JKutscha:
> 
> @echebbi @JKutscha what do you think?

Looks good to me 👍

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/utils/PitclipseUiUtils.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Executes the {@link Runnable} and in case of {@link SWTError} shows a dialog
+     * with the specified error message, using the specified parent..
+     * 
+     * @param runnable
+     * @param parent
+     * @param errorMessage
+     */
+    public static void safeExecute(Runnable runnable, Composite parent, String errorMessage) {
```
echebbi, line 67: Since it's a function I'd rather name it `executeSafely` (and the same goes for the other functions). What do you think?

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
```diff
@@ -33,20 +32,18 @@
...
+                browser = new Browser(parent, SWT.NONE);
+                browser.addProgressListener(new PitUiUpdatePublisher(browser));
+            },
+            parent,
+            "Browser cannot be initialized.");
```
echebbi, line 41: I'd like the popup to explicitely tell that the error is related to the summary view, could we say something like "Failed to create the Pit Summary view: browser cannot be initialized" or would it be too long?

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/utils/PitclipseUiUtils.java
```diff
@@ -0,0 +1,77 @@
...
+        try {
+            runnable.run();
+        } catch (SWTError e) {
+            MessageBox messageBox = new MessageBox(parent.getShell(), SWT.ICON_ERROR | SWT.OK);
+            messageBox.setMessage(errorMessage);
```
echebbi, line 72: Shouldn't we also display details about the error? Not sure about the stacktrace but I think that `e`'s message could help the user to fix/understand the issue.

echebbi (commented):
Some comments/feedbacks

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
LorenzoBettini: Just to separate better visually the 3 arguments

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/utils/PitclipseUiUtils.java
LorenzoBettini: You mean also `safelySetFocus` and `safelyDispose`, right?
They'd sound good to me, you're right!

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/view/PitView.java
LorenzoBettini: No, I think it would be fine!

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/utils/PitclipseUiUtils.java
LorenzoBettini: But, as you said here https://github.com/pitest/pitclipse/pull/150#discussion_r680315622 , the `SWTError` in such a case contains no message, doesn't it?

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/utils/PitclipseUiUtils.java
echebbi: Yes, I'd rename them `setFocusSafely` and `disposeSafely`

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/utils/PitclipseUiUtils.java
echebbi: It's the code snippet that made you think so? I apologize if my comment was misleading, browsing the sources from the IDE seems to indicate that `Browser` does set an error message:

```java
String errMsg = " because there is no underlying browser available.\n";
switch (SWT.getPlatform()) {
    case "gtk":
        errMsg = errMsg + "Please ensure that WebKit with its GTK 3.x bindings is installed (WebKit2 API level is preferred)."
                + " Additionally, please note that GTK4 does not currently have Browser support.\n";
        break;
    case "cocoa":
        errMsg = errMsg + "SWT failed to load the WebKit library.\n";
        break;
    case "win32":
        errMsg = errMsg + "SWT uses either IE or WebKit. Either the SWT.WEBKIT flag is passed and the WebKit library was not "
                + "loaded properly by SWT, or SWT failed to load IE.\n";
        break;
    default:
        break;
}
SWT.error (SWT.ERROR_NO_HANDLES, null, errMsg);
```
Even if there was no message, in the future this function may execute other runnables which may throw a `SWTError` that holds a message. Maybe we can check whether a message is available? Something like:
```java
} catch (SWTError e) {
    if (e.getMessage() != null) {
        errorMessage += "\n" + e.getMessage();
    }
    MessageBox messageBox = ...
    messageBox.setMessage(errorMessage);
}

Comment (LorenzoBettini) [reply bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/utils/PitclipseUiUtils.java:76]: OK, I pushed other commits to address your comments.

In particular, for this comment, I simulated an error like that

```java
    @Override
    public synchronized void createPartControl(Composite parent) {
        PitclipseUiUtils.executeSafely
            (() -> {
                browser = new Browser(parent, SWT.NONE);
                browser.addProgressListener(new PitUiUpdatePublisher(browser));
                throw new SWTError("Example\nAnother line");
            },
            parent,
            "Failed to create the Pit Summary view: browser cannot be initialized.");
    }
```

and I get this message dialog that looks OK, doesn't it?

image

LorenzoBettini:
@echebbi do you think we can merge now?

bundles/org.pitest.pitclipse.ui/src/org/pitest/pitclipse/ui/utils/PitclipseUiUtils.java
echebbi: Looks good, but is "Exit" a relevant title?

LorenzoBettini: It was actually there in the original code, but I agree it sounds strange... any suggestion?

echebbi: I think I'd create a `title` parameter and set the argument to "Failed to create the Pit Summary view":

```java
PitclipseUiUtils.executeSafely
    (() -> {
        ...
    },
    parent,
    "Failed to create the Pit Summary view",  // popup's title
    "Browser cannot be initialized");          // popup's body message
```
If that looks too cumbersome I'd replace "Exit" with "An error occurred", or some similar generic message.

LorenzoBettini: Perfect! Here's the result (I've already pushed the additional commit)

image

echebbi (approved):

PR #175, opened 2021-08-09
Make tests consistent

JKutscha:
* don't rerun pit, if we run in a timeout
* make sure we are getting the menu from the workbench window

### DIFF
--- a/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/RefactorMenu.java
+++ b/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/RefactorMenu.java
@@ -39,7 +39,7 @@ public class RefactorMenu {
 
     private void clickRename() {
         SWTBotMenuHelper menuHelper = new SWTBotMenuHelper();
-        menuHelper.findMenu(bot.menu(REFACTOR), RENAME).click();
+        menuHelper.findMenu(menuHelper.findWorkbenchMenu(bot, REFACTOR), RENAME).click();
     }
 
     public void renamePackage(String packageName) {
--- a/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitResultNotifier.java
+++ b/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitResultNotifier.java
@@ -138,7 +138,7 @@ public class PitResultNotifier implements ResultNotifier<PitResults> {
                 public String getFailureMessage() {
                     return "No summary was generated after the specified time.";
                 }
-            }, (timeOut > 0) ? timeOut : SWTBotPreferences.TIMEOUT);
+            }, (timeOut > 0) ? timeOut : Long.parseLong(System.getProperty(SWTBotPreferences.KEY_TIMEOUT)));
         }
 
         private void setSummary(Summary summary) {
--- a/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitResultNotifier.java
+++ b/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitResultNotifier.java
@@ -138,7 +138,7 @@ public class PitResultNotifier implements ResultNotifier<PitResults> {
                 public String getFailureMessage() {
                     return "No summary was generated after the specified time.";
                 }
-            }, (timeOut > 0) ? timeOut : Long.parseLong(System.getProperty(SWTBotPreferences.KEY_TIMEOUT)));
+            }, (timeOut > 0) ? timeOut : SWTBotPreferences.TIMEOUT);
         }
 
         private void setSummary(Summary summary) {
--- a/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/SWTBotMenuHelper.java
+++ b/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/SWTBotMenuHelper.java
@@ -16,32 +16,20 @@
 
 package org.pitest.pitclipse.ui.swtbot;
 
-import static org.junit.Assert.fail;
-
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
 import org.eclipse.swt.widgets.Menu;
 import org.eclipse.swt.widgets.MenuItem;
+import org.eclipse.swt.widgets.Shell;
 import org.eclipse.swtbot.eclipse.finder.SWTWorkbenchBot;
 import org.eclipse.swtbot.swt.finder.exceptions.WidgetNotFoundException;
 import org.eclipse.swtbot.swt.finder.finders.UIThreadRunnable;
+import org.eclipse.swtbot.swt.finder.results.Result;
 import org.eclipse.swtbot.swt.finder.results.WidgetResult;
 import org.eclipse.swtbot.swt.finder.widgets.SWTBotMenu;
 import org.eclipse.swtbot.swt.finder.widgets.SWTBotShell;
+import org.eclipse.ui.IWorkbenchWindow;
+import org.eclipse.ui.PlatformUI;
 
 public class SWTBotMenuHelper {
-    /**
-     * Text which identifies the workspace shell, to get workspace shell reliable
-     * with the swt bot while running local tests.
-     */
-    private static final String WORKSPACE_SHELL_TEXT_LOCAL = "junit-workspace";
-    /**
-     * Text which identifies the workspace shell, to get workspace shell reliable
-     * with the swt bot while running maven.
-     */
-    private static final String WORKSPACE_SHELL_TEXT_MAVEN = "data";
-
     private static final class MenuFinder implements WidgetResult<MenuItem> {
         private final SWTBotMenu parentMenu;
         private final String searchString;
@@ -83,15 +71,19 @@ public class SWTBotMenuHelper {
     }
 
     public SWTBotMenu findWorkbenchMenu(final SWTWorkbenchBot bot, final String menuString) {
-        for (SWTBotShell shell : bot.shells()) {
-            if (shell.getText().contains(WORKSPACE_SHELL_TEXT_LOCAL)
-                    || shell.getText().contains(WORKSPACE_SHELL_TEXT_MAVEN)) {
-                return shell.menu().menu(menuString);
-            }
-        }
+        return (new SWTBotShell(getActiveWorkbenchWindowShell(bot))).menu().menu(menuString);
+    }
 
-        fail("Could not find workbench shell.\n" +
-                "Shells found were: " + Stream.of(bot.shells()).map(s -> s.getText()).collect(Collectors.toList()));
-        return null; // never reached
+    private IWorkbenchWindow getActiveWorkbenchWindow(SWTWorkbenchBot bot) {
+        return UIThreadRunnable.syncExec(bot.getDisplay(), new Result<IWorkbenchWindow>() {
+            @Override
+            public IWorkbenchWindow run() {
+                return PlatformUI.getWorkbench().getActiveWorkbenchWindow();
+            }
+        });
+    }
+
+    private Shell getActiveWorkbenchWindowShell(SWTWorkbenchBot bot) {
+        return getActiveWorkbenchWindow(bot).getShell();
     }
 }
--- a/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/SWTBotMenuHelper.java
+++ b/tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/SWTBotMenuHelper.java
@@ -70,6 +70,14 @@ public class SWTBotMenuHelper {
         }
     }
 
+    /**
+     * This method does not rely on the focus and gets the workbench shell and from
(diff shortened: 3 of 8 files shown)


JKutscha:
@LorenzoBettini or @echebbi would you have a look at these changes and merge please?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/RefactorMenu.java
```diff
@@ -39,7 +39,7 @@ public void renameClass(String newClassName) {
 
     private void clickRename() {
         SWTBotMenuHelper menuHelper = new SWTBotMenuHelper();
-        menuHelper.findMenu(bot.menu(REFACTOR), RENAME).click();
+        menuHelper.findMenu(menuHelper.findWorkbenchMenu(bot, REFACTOR), RENAME).click();
```
LorenzoBettini, line 42: I need to understand: why is that required? The situation where this is required looks like when there's another dialog open? Doesn't this represent a bug in the test?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/SWTBotMenuHelper.java
```diff
@@ -64,4 +81,17 @@ public SWTBotMenu findMenu(final SWTBotMenu parentMenu,
             return new SWTBotMenu(menuItem);
         }
     }
+
+    public SWTBotMenu findWorkbenchMenu(final SWTWorkbenchBot bot, final String menuString) {
+        for (SWTBotShell shell : bot.shells()) {
```
LorenzoBettini, line 86: If I understand correctly you want to make sure the Eclipse window is the one that's active right?
If so, SWTBot FAQ shows a way to do that, which is easier than this, and does not require to distinguish between Eclipse and Maven build (https://wiki.eclipse.org/SWTBot/Troubleshooting) something like this should force the focus on the Eclipse workbench:

```java
UIThreadRunnable.syncExec(new VoidResult() {
  public void run() {
    PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell().forceActive();
  }
});
```

LorenzoBettini (changes requested):
I have a few doubts about the code: I think a few parts should be simplified as shown in the comments. I'd also like to know more about the situations they try to fix

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitMutationsViewTest.java
```diff
@@ -33,16 +34,22 @@ public static void setupJavaProject() throws CoreException {
...
-        "SURVIVED    | " + TEST_PROJECT + " | foo.bar | foo.bar.Foo |    7 | negated conditional");
-        pitclipseSteps.doubleClickMutationInMutationsView(mutation);
-        pitclipseSteps.mutationIsOpened(FOO_CLASS + ".java", 7);
-        mutation = fromMutationLine(
-        "SURVIVED    | " + TEST_PROJECT + " | foo.bar | foo.bar.Bar |    7 | negated conditional");
-        pitclipseSteps.doubleClickMutationInMutationsView(mutation);
-        pitclipseSteps.mutationIsOpened(BAR_CLASS + ".java", 7);
+        final long timeOutBefore = SWTBotPreferences.TIMEOUT;
+        try {
+            SWTBotPreferences.TIMEOUT = 10000;
```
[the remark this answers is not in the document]
JKutscha: Because this timeout value is used for how long we are waiting for the pit run. Currently, the timeout is set to 5 Seconds and I turned it up to 10 seconds to give pit enough time to finish.

Should I use the search timeout instead, but this seems to be very long?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/RefactorMenu.java
JKutscha: This is technically not required in this particular case, but this implementation doesn't rely on the focus, which in some cases depends heavily on the window manager of the OS.

With this implementation, we just make sure to always use the correct shell to open the menu.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/SWTBotMenuHelper.java
JKutscha: I don't want to use the focus as much as possible, because this can be a root of problems, as seen in #152.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitMutationsViewTest.java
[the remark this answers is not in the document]
LorenzoBettini: I think that we should use the same timeout consistently and make it configurable from the outside as we do right now... @echebbi what do you think?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/behaviours/pageobjects/RefactorMenu.java
LorenzoBettini: Could you please describe the scenario where this is required? I seem to recall that you propose this PR because otherwise, a few tests in another PR were failing? Or did I not recall correctly?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/SWTBotMenuHelper.java
LorenzoBettini: OK, but if you need to access the workbench shell at least do that with the Eclipse API and not based on the name of the window (which is different in Eclipse and in Maven). Something similar to what I showed above. For example, this is taken from `org.eclipse.swtbot.eclipse.finder.DefaultWorkbench.getActiveWorkbenchWindowShell()` (Once you have a `Shell` you can construct an `SWTBotShell`):

```java
	private IWorkbenchWindow getActiveWorkbenchWindow() {
		return UIThreadRunnable.syncExec(bot.getDisplay(), new Result<IWorkbenchWindow>() {
			@Override
			public IWorkbenchWindow run() {
				return PlatformUI.getWorkbench().getActiveWorkbenchWindow();
			}
		});
	}

	private Widget getActiveWorkbenchWindowShell() {
		return getActiveWorkbenchWindow().getShell();
	}
```

you can see that the workbench window can be accessed programmatically independently from the window text.

LorenzoBettini: Please, also have a look at https://github.com/pitest/pitclipse/pull/152#issuecomment-896862926
from my experience when I get the test failure finding the menu is because SWTBot does not find the shell active (typically the Eclipse workbench) that's why I'm forcing the activation of the workbench.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitMutationsViewTest.java
```diff
@@ -37,11 +37,11 @@ public void selectMutationOpensTheClassAtTheRightLineNumber() throws CoreExcepti
         coverageReportGenerated(2, 80, 0, 6, 0);
         PitclipseSteps pitclipseSteps = new PitclipseSteps();
         PitMutation mutation = fromMutationLine(
-        "SURVIVED    | " + TEST_PROJECT + " | foo.bar | foo.bar.Foo |    7 | negated conditional");
+                "SURVIVED    | " + TEST_PROJECT + " | foo.bar | foo.bar.Foo |    7 | negated conditional");
         pitclipseSteps.doubleClickMutationInMutationsView(mutation);
         pitclipseSteps.mutationIsOpened(FOO_CLASS + ".java", 7);
         mutation = fromMutationLine(
-        "SURVIVED    | " + TEST_PROJECT + " | foo.bar | foo.bar.Bar |    7 | negated conditional");
+                "SURVIVED    | " + TEST_PROJECT + " | foo.bar | foo.bar.Bar |    7 | negated conditional");
```
LorenzoBettini, line 44: are these changed indentations intentional?

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitResultNotifier.java
```diff
@@ -138,7 +138,7 @@ public boolean test() {
                 public String getFailureMessage() {
                     return "No summary was generated after the specified time.";
                 }
-            }, (timeOut > 0) ? timeOut : SWTBotPreferences.TIMEOUT);
+            }, (timeOut > 0) ? timeOut : Long.parseLong(System.getProperty(SWTBotPreferences.KEY_TIMEOUT)));
```
LorenzoBettini, line 141: In SWTBot you have

```java
public static long TIMEOUT = toLong(System.getProperty(KEY_TIMEOUT, "5000"), 5000);
```

so please, just rever this change, which is only confusing and completely equivalent

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/SWTBotMenuHelper.java
```diff
@@ -64,4 +69,21 @@ public SWTBotMenu findMenu(final SWTBotMenu parentMenu,
             return new SWTBotMenu(menuItem);
         }
     }
+
+    public SWTBotMenu findWorkbenchMenu(final SWTWorkbenchBot bot, final String menuString) {
```
LorenzoBettini, line 73: It might be the case to write a well-documenting Javadoc here, in particular, the problems it fixes

LorenzoBettini (changes requested):

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitMutationsViewTest.java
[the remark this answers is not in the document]
echebbi: > I think that we should use the same timeout consistently and make it configurable from the outside as we do right now

@LorenzoBettini I'm not sure I understood you, what do you mean exactly by "make it configurable from the outside"? As you said above the timeout is already configurable through the `org.eclipse.swtbot.search.timeout` property.

I agree that we should keep the timeout as consistent as possible thoughout the tests and avoiding changing it programmatically. Ideally I'd actually like to have a dedicated property that defines the timeout before we stop to wait for PIT so that we can change it without impacting the timeout for UI elements; something like `org.pitest.pitclipse.tests.pit.timeout`, but we'll see that it in another PR.

@JKutscha it looks like you reverted this change in your latest commit and the build seems to succeed so I guess this conversation can be closed? If you do need to increase the timeout, we set it [here](https://github.com/pitest/pitclipse/blob/157167faca6044b311f4a0606d0184770019ba45/tests/org.pitest.pitclipse.ui.tests/pom.xml#L89) for the Maven build:

https://github.com/pitest/pitclipse/blob/157167faca6044b311f4a0606d0184770019ba45/tests/org.pitest.pitclipse.ui.tests/pom.xml#L89

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/swtbot/PitResultNotifier.java
JKutscha: I did not know that.

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitMutationsViewTest.java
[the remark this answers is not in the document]
LorenzoBettini: > > I think that we should use the same timeout consistently and make it configurable from the outside as we do right now
> 
> @LorenzoBettini I'm not sure I understood you, what do you mean exactly by "make it configurable from the outside"? As you said above the timeout is already configurable through the `org.eclipse.swtbot.search.timeout` property.

Sorry for the misunderstanding: I meant "keep it" as it is right now :)

> 
> I agree that we should keep the timeout as consistent as possible thoughout the tests and avoiding changing it programmatically. Ideally I'd actually like to have a dedicated property that defines the timeout before we stop to wait for PIT so that we can change it without impacting the timeout for UI elements; something like `org.pitest.pitclipse.tests.pit.timeout`, but we'll see that it in another PR.

Yes, that should be trivial to implement. If you have time please open an issue or I'll do that ASAP

LorenzoBettini:
Thank you @JKutscha !

tests/org.pitest.pitclipse.ui.tests/src/org/pitest/pitclipse/ui/tests/PitclipsePitMutationsViewTest.java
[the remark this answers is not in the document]
echebbi: > If you have time please open an issue or I'll do that ASAP

Done, see #178

Issue #176, opened 2021-08-10
Cycle inside classpath while exporting


JKutscha:
I looked into exporting the plug in, but I was greeted with the following error:
![grafik](https://user-images.githubusercontent.com/50097340/128827276-134c6bf9-ef6a-4b21-a341-ea49b44f0624.png)

After looking at all pom.xml, MANIFEST.MF and build.properties files, I don't understand where this cycle is coming from. Especially because it seems to be a dependency from org.pitest to org.pitest.pitclipse.runner which should not even be possible.

Does anyone has an idea how this cycle could be resolved or found?

LorenzoBettini:
As we discussed in #177 we can close this

PR #177, opened 2021-08-10
Resolved dependency clash

JKutscha:
Fix for #176.

Soloution: Remove `bundle-version="2.0.0"` from MANIFEST.MF because it caused an extra plug in import inside the feature.xml.

### DIFF
--- a/bundles/org.pitest.pitclipse.listeners/META-INF/MANIFEST.MF
+++ b/bundles/org.pitest.pitclipse.listeners/META-INF/MANIFEST.MF
@@ -6,8 +6,8 @@ Bundle-Version: 2.1.2.qualifier
 Fragment-Host: org.pitest;bundle-version="1.6.8"
 Automatic-Module-Name: org.pitest.pitclipse.listeners
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
-Require-Bundle: org.pitest.pitclipse.runner;bundle-version="2.0.0"
 Bundle-ClassPath: .,
  META-INF/services/,
  src/
 Bundle-Vendor: Pitest.org
+Require-Bundle: org.pitest.pitclipse.runner
--- a/features/org.pitest.pitclipse.core.feature/feature.xml
+++ b/features/org.pitest.pitclipse.core.feature/feature.xml
@@ -228,6 +228,8 @@
    <requires>
       <import plugin="org.eclipse.core.runtime" version="3.11.1" match="compatible"/>
       <import plugin="org.eclipse.jdt.core" version="3.11.2" match="compatible"/>
+      <import plugin="org.pitest.pitclipse.runner" version="2.0.0" match="compatible"/>
+      <import plugin="org.pitest" version="1.6.8" match="greaterOrEqual"/>
       <import plugin="org.eclipse.core.resources" version="3.10.1" match="compatible"/>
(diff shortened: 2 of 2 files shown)


bundles/org.pitest.pitclipse.listeners/META-INF/MANIFEST.MF
```diff
@@ -6,8 +6,8 @@ Bundle-Version: 2.1.2.qualifier
 Fragment-Host: org.pitest;bundle-version="1.6.8"
 Automatic-Module-Name: org.pitest.pitclipse.listeners
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
-Require-Bundle: org.pitest.pitclipse.runner;bundle-version="2.0.0"
 Bundle-ClassPath: .,
  META-INF/services/,
  src/
 Bundle-Vendor: Pitest.org
+Require-Bundle: org.pitest.pitclipse.runner
```
LorenzoBettini, line 13: If this fixes the problem you experienced then I guess it's a bug in the mechanism you've used. Could you please tell me what you were trying to do when you got the cycle (which of course does not exist)? Did you try to create an update site from Eclipse?

features/org.pitest.pitclipse.core.feature/feature.xml
```diff
@@ -228,6 +228,8 @@
    <requires>
       <import plugin="org.eclipse.core.runtime" version="3.11.1" match="compatible"/>
       <import plugin="org.eclipse.jdt.core" version="3.11.2" match="compatible"/>
+      <import plugin="org.pitest.pitclipse.runner" version="2.0.0" match="compatible"/>
```
LorenzoBettini, line 231: these two are not required in the `<requires>` section since they are included as plugins or transitively through the included features of ours

LorenzoBettini (changes requested):
I have a few doubts about this fix

features/org.pitest.pitclipse.core.feature/feature.xml
JKutscha: If so, the calculation of the Eclipse PDE is not working properly.

My steps to create this were:
1. Open this file in Eclipse
2. Switch to the dependency tab of the feature.xml
3. Click on compute
grafik

bundles/org.pitest.pitclipse.listeners/META-INF/MANIFEST.MF
JKutscha: I did try to create a local update site of this plug in.

This was required for me to resolve the issue.

JKutscha:
This seems not to fix the issue after a restart of eclipse, which I didn't test before.

features/org.pitest.pitclipse.core.feature/feature.xml
LorenzoBettini: Probably it finds those dependencies and it adds them but they're useless and probably dangerous. Usually I never used that mechanism because it finds compile dependencies that are implicitly and automatically taken into consideration by P2 itself. This list should be used to specify runtime dependencies that cannot be computed automatically

bundles/org.pitest.pitclipse.listeners/META-INF/MANIFEST.MF
LorenzoBettini: To that aim it's best to use the Maven Tycho build

PR #181, opened 2021-08-24
Correcting the default init from preferences

JKutscha:
The current init of the preferences is broken and writes a wrong value in the mutators preference.

### DIFF
--- a/bundles/org.pitest.pitclipse.core/src/org/pitest/pitclipse/core/preferences/PreferenceInitializer.java
+++ b/bundles/org.pitest.pitclipse.core/src/org/pitest/pitclipse/core/preferences/PreferenceInitializer.java
@@ -16,15 +16,11 @@
 
 package org.pitest.pitclipse.core.preferences;
 
-import org.eclipse.core.runtime.preferences.AbstractPreferenceInitializer;
-import org.eclipse.jface.preference.IPreferenceStore;
-import org.pitest.pitclipse.core.PitCoreActivator;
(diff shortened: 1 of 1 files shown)


LorenzoBettini:
That's something I had spotted in the past myself but never fixed that because I couldn't figure out how the bug actually revealed itself. How did you find the wrong behavior?

JKutscha:
I found this bug, after resetting my test workspace.

LorenzoBettini:
> I found this bug, after resetting my test workspace.

and what's the value it writes?

JKutscha:
> 
> and what's the value it writes?

I‘m not sure, if I understand. What value are you referring to?

LorenzoBettini:
> > and what's the value it writes?
> 
> I‘m not sure, if I understand. What value are you referring to?

I mean, how have you detected that it was setting the wrong value?
When I opened my fresh workspace it was using the DEFAULT mutators (but I'm talking about the current released version, not the one under development), that's why when I spotted that strange setting I did not think about a bug since the behavior, in the end, was the expected one...

JKutscha:
> 
> I mean, how have you detected that it was setting the wrong value?
> When I opened my fresh workspace it was using the DEFAULT mutators (but I'm talking about the current released version, not the one under development), that's why when I spotted that strange setting I did not think about a bug since the behavior, in the end, was the expected one...

The value was set to "defaultMutators" instead of the DEFAULTS.
I think this is the result of the rework I did on the preferences and the mutators itself.

LorenzoBettini:
OK, merged! Thanks!

Issue #157, opened 2021-07-23
Pitclipse creates new LaunchConfiguration even if launch config exists


JKutscha:
I am currently working on tests for #114 and I want to manipulate a run configuration to run the tests.

The first run uses the correct config, but the next create new configurations and cause the tests to fail.

I am not sure why this is happening and when, but I noticed it even using pitclipse by hand.

echebbi:
Could you detail exactly the steps so that we can reproduce the issue?

If you're talking about running Pit from the contextual menu (right-click on a project > Run As > Pit) then there's indeed an odd behavior. If a run configuration already exists then a dialog opens to ask the user to select an existing configuration. However, even if the user selects an existing configuration, a new one is always created.

If that's the issue you're talking about then its root lies here:

https://github.com/pitest/pitclipse/blob/70a1613a6f344fe3ea73605a12bf658deae89797/bundles/org.pitest.pitclipse.launch.ui/src/org/pitest/pitclipse/launch/ui/PitLaunchShortcut.java#L138

I _guess_ that the author of the code was expecting the new configuration `tmp` to only be saved when `existingConfig` is an empty Optional (which means that the user hasn't selected any configuration). However, in Java, arguments are passed by value, hence `tmp.doSave()` is always called.

Can be fixed with a lambda:

```java
ILaunchConfiguration config = existingConfig.orElseGet(tmp::doSave);
```

JKutscha:
> If you're talking about running Pit from the contextual menu (right-click on a project > Run As > Pit) then there's indeed an odd behavior. […]

This is the behaviour I meant.

> Can be fixed with a lambda:
> 
> ```java
> ILaunchConfiguration config = existingConfig.orElseGet(tmp::doSave);
> ```

Good to know, thanks :)

PR #184, opened 2021-09-03
[Bug Fix] No longer create run config, if one exists

JKutscha:
Fix for #157.
Thanks to @echebbi.

### DIFF
--- a/bundles/org.pitest.pitclipse.launch.ui/src/org/pitest/pitclipse/launch/ui/PitLaunchShortcut.java
+++ b/bundles/org.pitest.pitclipse.launch.ui/src/org/pitest/pitclipse/launch/ui/PitLaunchShortcut.java
@@ -135,14 +135,12 @@ public class PitLaunchShortcut implements ILaunchShortcut2 {
     private void performLaunch(IJavaElement element, String mode) throws InterruptedException, CoreException {
         ILaunchConfigurationWorkingCopy tmp = createLaunchConfiguration(element);
         Optional<ILaunchConfiguration> existingConfig = findExistingLaunchConfiguration(tmp);
-        ILaunchConfiguration config = existingConfig.orElseGet(() -> {
-            try {
-                return tmp.doSave();
-            } catch (CoreException e) {
-                // produces a NullPointerException
-                return null;
-            }
-        });
+        ILaunchConfiguration config;
+        if (existingConfig.isPresent()) {
+            config = existingConfig.get();
+        } else {
+            config = tmp.doSave();
+        }
         DebugUITools.launch(config, mode);
     }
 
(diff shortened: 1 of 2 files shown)


bundles/org.pitest.pitclipse.launch.ui/src/org/pitest/pitclipse/launch/ui/PitLaunchShortcut.java
```diff
@@ -135,7 +135,14 @@ private void showNoTestsFoundDialog() {
     private void performLaunch(IJavaElement element, String mode) throws InterruptedException, CoreException {
         ILaunchConfigurationWorkingCopy tmp = createLaunchConfiguration(element);
         Optional<ILaunchConfiguration> existingConfig = findExistingLaunchConfiguration(tmp);
-        ILaunchConfiguration config = existingConfig.orElse(tmp.doSave());
+        ILaunchConfiguration config = existingConfig.orElseGet(() -> {
+            try {
+                return tmp.doSave();
+            } catch (CoreException e) {
```
LorenzoBettini, line 141: Unfortunately, Eclipse API is not "lambda ready" due to checked exceptions like CoreException, and that case is not covered (as new code) https://sonarcloud.io/component_measures?id=org.pitest%3Aorg.pitest.pitclipse&metric=new_coverage&pullRequest=185&selected=org.pitest%3Aorg.pitest.pitclipse%3Abundles%2Forg.pitest.pitclipse.launch.ui%2Fsrc%2Forg%2Fpitest%2Fpitclipse%2Flaunch%2Fui%2FPitLaunchShortcut.java&view=list

Besides not covered I think this solution is bad (that is, returning null) because

- it will generate an NPE in that case (as you write in the comment)
- we don't benefit from the `throws` declaration that already takes care of propagating such an exception

For these reasons, I'd suggest using the "old school" style, instead of a lambda, e.g., something like `if (config.isPresent()).

LorenzoBettini (changes requested):
Besides the comment in the code, I'm a bit worried that there's no test specific for this change. The code is covered, but maybe it would be better to have a specific test for this change (possibly making sure that it fails without this change). If I remember correctly in our test utility code there's a part that was taking care of interacting with a dialog popping up because of multiple run configurations for the same run (that was due to avoid the side effects of this bug, which was probably unknown that time?). Maybe one could get rid of that code and write a test that simply tries to run the same "thing" twice: the second time it should succeed even without having to interact with that dialog (without your change that test should fail because the popping dialog would disturb the test I guess).

@echebbi what do you think?

echebbi:
> @echebbi what do you think?

I agree. Even small bug fixes like this one should introduce new tests that ensure the bug doesn't show up again in the future.

LorenzoBettini (approved):
Sorry it took so long :)
I'm going to merge this now

Issue #209, opened 2022-08-31
Switch to AssertJ instead of Hamcrest


LorenzoBettini:
AssertJ is much more readable and fluent

PR #212, opened 2022-09-19
fix deprecations

Bananeweizen:


Bananeweizen:
This is confusingly inconsistent:
* The current target platform uses 2022-06, thereby leading to deprecation warnings.
* The Github build uses an older target platform, which seems not to be used in the normal Maven build, thereby making local and CI build different.
* The Orbit version does not match the Eclipse platform version in the target (they must both be from the same date)
* The older target uses Photon, but the marketplace entry says compatible with Oxygen (which is not tested anywhere, not even with the older.target)
* The requirements in the bundles and features don't fit the target platform restrictions, thereby leading to "can install, but doesn't run" issues for some Eclipse versions.

Any idea what's really the smallest Eclipse version meant to be supported?

LorenzoBettini:
> This is confusingly inconsistent:
> 
> * The current target platform uses 2022-06, thereby leading to deprecation warnings.

Yes, I know. Note that there's another one besides the ones you address here (I seem to remember getClassPath in some of our launch configuration classes) that we cannot yet fix because the JUnit launcher itself from JDT still relies on that method that we override. 

> * The Github build uses an older target platform, which seems not to be used in the normal Maven build, thereby making local and CI build different.

We have a dedicated job to verify that we can still build and test against the oldest TP that we want to support, in this case Photon. Of course, Pitclipse might also be installed in older platforms, but we don't test such situations.

> * The Orbit version does not match the Eclipse platform version in the target (they must both be from the same date)

I can't do that right now because that would break some of our unit tests, from what I understand, due to some incompatibilities between Mockito and Hamcrest. I did not have time to investigate further, and for the moment, it's safe to use the older Orbit, since the things we take from Orbit are now used only for testing purposes (since this is solved https://github.com/pitest/pitclipse/issues/136). When also https://github.com/pitest/pitclipse/issues/209 and especially https://github.com/pitest/pitclipse/issues/209 will be solved, I can basically remove most of things from Orbit and I can focus on updating Orbit (again, at that time, we will still use it only for testing).

> * The older target uses Photon, but the marketplace entry says compatible with Oxygen (which is not tested anywhere, not even with the older.target)

Right! I've updated the MarketPlace entry!

> * The requirements in the bundles and features don't fit the target platform restrictions, thereby leading to "can install, but doesn't run" issues for some Eclipse versions.

mh... which ones please? I'd like the minor versions of dependencies to reflect Photon.

> 
> Any idea what's really the smallest Eclipse version meant to be supported?

As I said, it's currently Photon. @echebbi prefers to support at least that version, unless he changed his mind ;)

